Issue #10, opened 2022-09-26
Monitor and log website traffic


ctmbl:
**Context:**
The #2 PR from @atxr has introduced User Authentication to our website: a mandatory feature to develop other services around the website. However this PR also brought many security concerns.. and we, as students, aren't fully capable of designing an entirely safe website, even a small one, despite all our interest towards cyber-security. At least that's what I think.

**Problem:**
Even with our best effort we can't design a safe website.

**Solution:**
I then want to propose a complementary solution, that I think is also widely used in the industry and is quite an interesting challenge.
Because we can't guarantee that the website is safe we could monitor the traffic and log it. That would allow to detect as early as possible breaking-in attempts.
I don't know what kind of security and monitoring the current website and framework offers (defense against brute-forcing, automatic log to a file etc) but I'm interested in looking deeper into the subject!

amtoine:
great idea!

will probably not have time to help, but looking forward to it :+1:

atxr:
That's an amazing idea. I didn't implemented any logs at the moment for the backend, and I don't think node does it automatically.
Do you want to add it to the v0.0.2 milestone or maybe another one? It doesn't seems to be related to the changes of v0.0.2

atxr:
Some useful links:
- https://stackoverflow.com/questions/10815218/where-is-nodejs-log-file
- https://blog.appsignal.com/2021/09/01/best-practices-for-logging-in-nodejs.html
- [winston package](https://www.npmjs.com/package/winston)
- [log4js package](https://github.com/log4js-node/log4js-node) (yes you read well, the famous one)(version js)
-

ctmbl:
Great! I can look at these subject, I have a (little) bit of logging experience since my internship of the past summer so even if it's a new language and framework I can look into it :wink:

atxr:
Thank you! Don't hesitate if you have questions about the framework!
I assigned this feature to the v0.0.2 milestone!

ctmbl:
I just discovered that `nginx` already logs every connection attempt, maybe we should build or configure on top of it a tool to monitor it?
I don't think there is traffic that isn't logged by nginx given that any attempt from a browser to reach the backend will eventually pass by nginx? but @atxr I'd like your advices

see `sudo docker logs iscscfr-nginx-1 | tac | less` on the iscsc remote server

ctmbl:
will wait for #86 and #95

PR #17, opened 2022-09-29
Close #14 : Unhandled Error when trying to upload an article that is too big

### LINKED ISSUE #14: Unhandled Error when trying to upload an article that is too big
Trying to upload an article that is too big causes an error in the backend; I produced this error by trying to upload a 64ko file as a markdown article.

The preview worked fine, but pressing the "Add article" button creates the following error (see end of message) in the **backend**. 
From the user's perspective, nothing happened as no new article is created, and no error is reported.

**Details:**
- OS: Windows 10
- npm version: 8.19.1
- node version: 18.9.1

Error message:
```
PayloadTooLargeError: request entity too large
    at readStream (C:\Users\*****\Documents\repo\site\backend\node_modules\raw-body\index.js:156:17)
    at getRawBody (C:\Users\*****\Documents\repo\site\backend\node_modules\raw-body\index.js:109:12)
    at read (C:\Users\*****\Documents\repo\site\backend\node_modules\body-parser\lib\read.js:79:3)
    at jsonParser (C:\Users\*****\Documents\repo\site\backend\node_modules\body-parser\lib\types\json.js:135:5)
    at Layer.handle [as handle_request] (C:\Users\*****\Documents\repo\site\backend\node_modules\express\lib\router\layer.js:95:5)
    at trim_prefix (C:\Users\*****\Documents\repo\site\backend\node_modules\express\lib\router\index.js:328:13)
    at C:\Users\*****\Documents\repo\site\backend\node_modules\express\lib\router\index.js:286:9
    at Function.process_params (C:\Users\*****\Documents\repo\site\backend\node_modules\express\lib\router\index.js:346:12)
    at next (C:\Users\*****\Documents\repo\site\backend\node_modules\express\lib\router\index.js:280:10)
    at expressInit (C:\Users\*****\Documents\repo\site\backend\node_modules\express\lib\middleware\init.js:40:5)
```

atxr:
I propose a fix for #14 
It's just a maximum file size for the upload. The maximum size in 40kb at the moment, because I didn't have any idea. But I created a config file for the frontend and this value can be modified fastly in a json file.


frontend/src/config/config.json
```diff
@@ -0,0 +1,3 @@
+{
+    "FILE_SIZE_MAX": 40000
```
gbrivady, line 2: I think we should mention with a comment that 40000 bytes is the **absolute maximum**, and instead put this value to like 32Kb or something

frontend/src/pages/blog/CreateArticle.js
```diff
@@ -51,15 +52,19 @@ const CreateArticle = () => {
...
-    file.text().then((t) => {
-      const lines = t.split("\n");
-      console.log(lines);
-      setTitle(lines[0]);
-      setSummary(lines[1]);
-      setBody(lines.slice(3).join("\n"));
-    });
+    console.log(file.size)
+    console.log(Config.FILE_SIZE_MAX)
+    if (file.size < Config.FILE_SIZE_MAX) {
```
gbrivady, line 57: And I think here, we should hard cap it to the max value of 40k bytes to be sure editing the config file don't cause a problem

frontend/src/pages/blog/CreateArticle.js
```diff
@@ -51,15 +52,19 @@ const CreateArticle = () => {
...
+        file.text().then((t) => {
+          const lines = t.split("\n");
+          setTitle(lines[0]);
+          setSummary(lines[1]);
+          setBody(lines.slice(3).join("\n"));
+        });
+    }
+    else {
+        setError("File is too large!")
```
gbrivady, line 66: Mention the max file size (ie `Config.FILE_SIZE_MAX` as a string) in the error message for user-friendliness

gbrivady (changes requested):
Tested that 40k bytes is the limit, and it seems to be the case; almost good for me but I think some little changes could be useful for user-friendliness, and so nothing breaks if someone messes with the config file.

frontend/src/config/config.json
atxr: I don't really understand :thinking: You want to put 32kb instead of 40kb?

frontend/src/pages/blog/CreateArticle.js
atxr: Again I don't get it. You want to hard code the maximum? So what's the point of the config file?

frontend/src/pages/blog/CreateArticle.js
atxr: Yes that's a good idea

frontend/src/config/config.json
ctmbl: I don't understand your point neither :confused: 
But a comment saying that tjis value is in byte would be good!

frontend/src/pages/blog/CreateArticle.js
```diff
@@ -51,15 +52,19 @@ const CreateArticle = () => {
...
-    console.log(file);
-    console.log(file.text());
-    file.text().then((t) => {
-      const lines = t.split("\n");
-      console.log(lines);
-      setTitle(lines[0]);
-      setSummary(lines[1]);
-      setBody(lines.slice(3).join("\n"));
-    });
+    console.log(file.size)
```
ctmbl, line 55: Is using console.log a good habit by the way ?
You could be a little bit explicit or do you prefer waiting for proper logging ?

ctmbl (changes requested):
A very slight change but that seems good!

frontend/src/pages/blog/CreateArticle.js
gbrivady: You are basically assuming that the config file is correct, but you should not for obvious reason.

The maximum (40Kb) is mostly there so no error happens, but as of right now if you were to put in the config file a maximum size of 300Kb for example, it would pass the test but still generates the error, effectively fixing nothing.

So the test should be something like `if (file.size < Config.FILE_SIZE_MAX && file.size < 40000)`.

And you may also want a different maximum size in the config file to limit space/ressource usage - I don't think a lot of people are going to type 30Kb of text.

frontend/src/config/config.json
gbrivady: We can gloss over that part, but I 100% think that it should mention in a commentary that 40k bytes is a **hard** maximum so no error happens on the backend. And I agree that saying it is in byte is a good ide

gbrivady (approved):

frontend/src/pages/blog/CreateArticle.js
atxr: The `console.log` is just logging for me when coding.
And yes the real logs will come with #10

frontend/src/pages/blog/CreateArticle.js
gbrivady: Since we had the conversation and the max value  before an error is commented somewhere,  the bare minimum is there, so I'm fine with closing this (and will do that).

gbrivady (approved):
Good enough to merge now.

atxr:
I added comments in the json config file in [c283b14](https://github.com/iScsc/iscsc.fr/pull/17/commits/c283b1493fa0bcf3c7d866f50b49b70d9a7ff979). It might look a bit weird, but their is no comments in json files.

ctmbl (approved):
It's a go :rocket:

atxr:
Finally after some discussion with @ctmbl we will cancel these changes
In this solution, I implemented a frontend solution do deal with the big files because I thought the error was from the frontend.
However the backtrace shows that the error is thrown in the backend.
I will then check the size on the backend instead of the frontend and remove this useless test

atxr:
The bug spotted by @gbrivady comes from the middleware `express.json`. This is a piece of code that runs each time between the frontend and backend and that parses the requests to json.
I raise the maximum size of this middleware to 1MB in [4652cf1](https://github.com/iScsc/iscsc.fr/pull/17/commits/4652cf1821c19452b3ff6cc73a108f42953ad217)
The goal is to limit errors from this middleware and to try to catch big file errors in the frontend/backend
For the frontend, nothing changed, same checks than before to deny big file reading.
For the backend, I added in [e1b47da](https://github.com/iScsc/iscsc.fr/pull/17/commits/e1b47da9f9e8b632a484d6cd227ca48d3f3be1eb) a bunch of tests performed before the backend creates a new article in the database
Let me know if it's clearer for you now!
I will ask you to check these changes before merging.
(I ping @ctmbl here to)

backend/app.js
```diff
@@ -7,18 +7,18 @@ const articleRoutes = require('./routes/articles');
...
-app.use('/api/users', userRoutes);
-app.use('/api/articles', articleRoutes);
-
 const {
     DB_USER,
     DB_PASSWORD,
     DB_NAME,
     PORT,
-    SALT_ROUND
+    BODY_MAX
```
ctmbl, line 15: Why did `SALT_ROUND` has been removed?

backend/controllers/articleController.js
```diff
@@ -41,12 +41,22 @@ const getById = async (req, res) => {
 
 const create = async (req, res) => {
   const { title, body, summary } = req.body;
+  console.log(body.length, process.env.MAX_BODY)
+  if (title.length > process.env.MAX_TITLE) {
+    return res.status(400).json({ error: `Your title should be less than ${process.env.MAX_TITLE} characters.` });
+  }
+  if (summary && summary.length > process.env.MAX_SUMMARY) {
+    return res.status(400).json({ error: `Your summary should be less than ${process.env.MAX_SUMMARY}300 characters.` });
```
ctmbl, line 49:
```suggestion
    return res.status(400).json({ error: `Your summary should be less than ${process.env.MAX_SUMMARY} characters.` });
```

backend/controllers/articleController.js
```diff
@@ -41,12 +41,22 @@ const getById = async (req, res) => {
 
 const create = async (req, res) => {
   const { title, body, summary } = req.body;
+  console.log(body.length, process.env.MAX_BODY)
+  if (title.length > process.env.MAX_TITLE) {
+    return res.status(400).json({ error: `Your title should be less than ${process.env.MAX_TITLE} characters.` });
+  }
+  if (summary && summary.length > process.env.MAX_SUMMARY) {
```
ctmbl, line 48: Why here testing if `summary` exists before checking length but not for `body`?

ctmbl (changes requested):
I have mainly two points:
 - try not to mess between bytes and characters! there are the same for ASCII but that's not true with UTF-8 or other UTF
 - I don't get the point of having 3 different size limits... 1MB for the middleware , 3kb for the backend and 40kb in the frontend??

backend/app.js
atxr: It's a useless line in this file, I didn't know how to clean it

backend/controllers/articleController.js
atxr: Title and body are `required` in the article model but summary is not

atxr:
> I have mainly two points:
>  - try not to mess between bytes and characters! there are the same for ASCII but that's not true with UTF-8 or other UTF

I know but in node I check the character length for simplicity and in the frontend the file size 

>  - I don't get the point of having 3 different size limits... 1MB for the middleware , 3kb for the backend and 40kb in the frontend?? 

The 3kb in the backend is an error it should be 40kb
And I don't really know how to deal with this. The middleware needs a file limit, otherwise it throws weird errors which are difficult to catch. I'll try something to catch them later, but at the moment, I put 1Mb in the middleware to trigger the backend/frontend error before
Moreover, the middleware deals with the whole request so body + title + summary + author + headers + other things maybe 
So its limit can be a bit higher.

backend/.env.example
```diff
@@ -14,3 +14,8 @@ SALT_ROUND=10  # number of rounds for salt generation. Default 10.
 # JWT
 JWT_SECRET=XXX  # random sercet to sign the JWT, NEED TO BE CHANGED FOR SECURITY REASONS
 JWT_EXPIRATION="3d"  # time before expiration of JWT, default 3 days
+
+# FILE UPLOAD
+MAX_TITLE=500  # max length of the article title, default 500b
+MAX_SUMMARY=500  # max length of the article summary, default 500b
+MAX_BODY=30000  # max length of the arcitle body, defautl 3kb
```
gbrivady, line 21: Good change! Was going to suggest a max size for typed in another issue, but it is also nice to address it right now

gbrivady (approved):
Minus one fix already mentionned by ctmbl, I think we are definetly good to go!

backend/controllers/articleController.js
atxr: Fix [205be45](https://github.com/iScsc/iscsc.fr/pull/17/commits/205be459073bf5f3dec191349f58d7771743748d)

atxr:
Alright! We need @ctmbl approval and we are good for closing this one!

backend/controllers/articleController.js
ctmbl: Ok tks!

ctmbl (approved):
> The 3kb in the backend is an error it should be 40kb

Apart from this point which hasn't been modifed @atxr I think it's good to go

atxr:
> > The 3kb in the backend is an error it should be 40kb
> 
> Apart from this point which hasn't been modifed @atxr I think it's good to go

I spotted why I did that. The size in the frontend is in byte and in the backend in char (don't ask why...)
So if it's utf-8 encoded, it's 4 bytes and 40kb is 10k chars. I fixed in [84ecb33](https://github.com/iScsc/iscsc.fr/pull/17/commits/84ecb338e99ee7a3f84f2116944e7df84ee3e8f4)
Should be ready to merge!
Sorry for the delay :rocket:

ctmbl:
@atxr 
Actually I think there is a slight misunderstanding here, even if I'm fine with 10k chars as a limit for he backend.

UTF-8 encoding hasn't a constant size in Bytes for char as you can read [on the wikipedia page](https://en.wikipedia.org/wiki/UTF-8) or in either of these StackOverflow answers: [1](https://stackoverflow.com/questions/19212306/whats-the-difference-between-ascii-and-unicode/41198513#41198513) [2](https://stackoverflow.com/questions/22404493/is-there-a-drastic-difference-between-utf-8-and-utf-16/22404874#22404874). By the way that's the reason why UTF-8 is so powerfull! It is ASCII transparent and then really light for common characters but can also with heavier characters (up to 4 Bytes  = 4Octets = 32 bits) encode less common characters while ASCII can't. UTF-32 on the other side is 4 Bytes constant for each character but isn't widely used because it's quite heavy...

Then I would say that you can't guess what will be the size in Bytes of 10k UTF-8 characters but 10k seems great because you can't set the perfect value :confused:

atxr:
Mmm yes indeed I already read about that I forgot :thinking: 
If we assume the worst case then, with only heavy 4 bytes characters, 10k chars = 40kb
So with this 10k chars limits we ensure at most 40kb of data :thinking:

atxr:
I put 40k everywhere after a private discussion with @ctmbl !

PR #19, opened 2022-10-01
Bump to version 0.1.0

ctmbl:
here is a PR for the first official version bump of the website.

Some questions must be address:
 - Where should the version go
   - :one: `frontend/package.json` and `backend/package.json` with same version (because for now there are developed at the same time)
   - :two: `frontend/package.json`, `backend/package.json` AND a root `package.json` indicating the version of the WHOLE website (the solution I chose in this PR :heart: )
 - What should we do about the wrong tag `v0.0.1` (I think we let it as it is and jump to next version)

The wiki will have be updated after this discussion
Right after this merging we will tag the commit with `v0.1.0`

**HURRY UP** this PR must be resolved before any other (especially #17) you have my point of view please give yours @atxr @amtoine


atxr:
The second method you propose is definitely the good one for long term.
I think soon the backend and frontend development will be separated (for the best)
We should still be able to track the version of each one and release a global version for the website when needed.
So it's ok for me!
Just I think you forgot to add the version in frontend/package.json

atxr:
Did you manage to find some documentation on removing an existing tag?

amtoine (approved):
the PR in itself looks good :+1:

amtoine:
you should be able to simply 
```bash
git tag --delete v0.0.1
git checkout 31e96b3
git tag v0.1.0 -m "blah blah"
git push --tags
```
:+1:

atxr (changes requested):
What about `frontend/package.json` ?

amtoine:
> What about `frontend/package.json` ?
>
in `nushell`, i get
```
nushell> [. frontend backend] | each {
  |it|
  open $"($it)/package.json" | get version
}
╭───┬───────╮
│ 0 │ 0.1.0 │
│ 1 │ 0.1.0 │
│ 2 │ 0.1.0 │
╰───┴───────╯
```
is there anything to do in `frontend/package.json`? :thinking:

atxr (approved):
Oh yes indeed, it's aleady the right version

PR #26, opened 2022-11-04
Fix bump version

ctmbl:
This PR aims more to be a look into what we maybe miss during the first version bump than a real addition to the project, it personally raised for me many question on how to rightly version a software and in the iscsc.fr case a website.


amtoine:
> This PR aims more to be a look into what we maybe miss during the first version bump than a real addition to the project, it personally raised for me many question on how to rightly version a software and in the iscsc.fr case a website.
>
i'm not sure what this is about then :thinking:

**maybe you want to make this a draft?**

ctmbl:
You're right it should be a DRAFT :wink: but you still can review the changes I made

backend/package.json
```diff
@@ -1,7 +1,8 @@
 {
-  "name": "server",
+  "name": "iscsc.fr-backend",
   "version": "0.1.0",
   "description": "Backend of the iscsc.fr website",
+  "private": true,
```
amtoine, line 5: simply not sure what `private` is supposed to do or enforce :yum:

amtoine (commented):
looks better :thumbsup:

backend/package.json
ctmbl: It simply forbid to publish package via `npm publish` https://docs.npmjs.com/cli/v6/configuring-npm/package-json#private

atxr:
Is it ready for review?

backend/package.json
amtoine: makes sense!

amtoine:
> Is it ready for review?
>
you marked it as such yourself right? :confused:
@ctmbl ?

ctmbl:
@atxr yes it is :wink:

ctmbl:
In particular I would really like any opinion on the [unused fields](https://docs.npmjs.com/cli/v6/configuring-npm/package-json/) of the `package.json` as well as `name` and `description`.

atxr:
I didn't get the point of this PR sorry can you explain it agan?

amtoine:
> In particular I would really like any opinion on the [unused fields](https://docs.npmjs.com/cli/v6/configuring-npm/package-json/) of the `package.json` as well as `name` and `description`.
>
maybe we just don't need them :thinking:

otherwise, you could set default values for each one of them and then we would be able to discuss these values

ctmbl:
@atxr 

> I didn't get the point of this PR sorry can you explain it again?

you're right I've not been clear enough here, in fact this PR has 2 points:
 - first, arise the question about the `package-lock.json`s (we talked about together in private) which hadn't been updated of the 1st version bump (you can check it with `git checkout v0.1.0 && cd backend && npm install && git diff`), **especially** the `./package-lock.json` that can look useless because we don't run `npm install` at root but what if we need to in the future? so here the questions: 
 **Should we run `npm install` when bumping version to update `package-lock.json`s? Should we run it at root? or include it to the `.gitignore`?** 
 - second, simply make the different `package.json` more consistent by adding/updating `name` and `description`

Sure these 2 things aren't of great importance :sweat_smile: but I'm a perfectionist I like things properly done :sweat_smile:

ctmbl:
@amtoine 
> maybe we just don't need them thinking

Absolutely fine to me but I had to ask!

atxr:
Ok it's clearer!
- For the first point, I think we should include the package-lock.json and switch every `npm install` to `npm ci`
However the package-lock.json in the root directory is useless because it will be empty (no packages are installed in root)
- for the second point, adding the name parameter looks good!

ctmbl:
@atxr 
ok so should I remove the root `package-lock.json`?

atxr:
> @atxr 
> ok so should I remove the root `package-lock.json`?

Yes I think so!

amtoine:
feel free to ping me with a new review when needed :+1:

ctmbl:
@atxr @amtoine 
`package-lock.json` has been removed I think we're able to merge this

amtoine (approved):
looks ok to me

atxr (approved):
That's better! Let's merge!

Issue #13, opened 2022-09-28
Automation and enhancement of the dev workflow


atxr:
In this issue, I would like to highlight the fact that the workflow of this project is not automated yet, and that a lot of tasks are still done by hand. To cite some of them:
- Sharing of `. env`
- Deployment of the website
- Test integration (cf #11)

Moreover, we can imagine a deployment of the website soon and this will induce a major change on the dev workflow. For example, developers mustn't play anymore with the official database used for the deployed version.

In order to avoid these issues, I would like to propose some basic scripts to make the development process easier. A first PR with some changes in how to deploy the website for dev purpose will be open soon.

amtoine:
definitely something to address!
100% agreed with what is said above :ok_hand:

another question that might fit in this issue: **how and when do we deploy the blog on https://iscsc.fr/?** :yum:

atxr:
> another question that might fit in this issue: **how and when do we deploy the blog on https://iscsc.fr/?** yum

I was thinking about opening another issue for this one, but go ahead if you want :wink:

ctmbl:
Alright you have a good point here @atxr
I can't wait to see your PR opened!

About the "messing with the DB" thing wouldn't it be possible to, during development process, redirect the data towards a different table of the DB to let the developers test everything without messing up with the running website? 

Linked to that point I have another request:
A **tag** of the version commit on the main branch. Now that v0.0.1 is released we should add a tag (I mean a **git tag**) on a commit of the main branch. But tagging a *PR commit* is not clean neither tagging an *empty*. 
I then propose a workflow I witnessed during my summer internship: after the last PR/bug fix we stop every merging on main and push a single commit for example: "Release v0.0.1" containing an update of a .json file for example:
```
{"version":"0.0.1"}
```
I find this quite clean and explicit! And just before, after or even in this commit we could also add the needed update to redirect the connection to a "Running" table of the DB (in contrast with a "Dev" during dev time).
Let me know if I'm not clear or if you want me to open another Issue for this.

ctmbl:
Absolutely!

amtoine:
that's cool :+1:

atxr:
> About the "messing with the DB" thing wouldn't it be possible to, during development process, redirect the data towards a different table of the DB to let the developers test everything without messing up with the running website?
> 

It's the idea yes! I'll create two tables `users-dev` and `articles-dev` alongside in the database

[...]

PR #29, opened 2022-11-10
FEATURE: add an env-sharing script

amtoine:
Related to #13.

This PR adds a script, in the `./scripts/` directory, to easily share secret files to the members of the `iscsc.fr` team.

### usage
```bash
./scripts/gpg-share.sh $file $[list of gpg users]
```
#### legend
:large_blue_circle: there is an info
:green_circle: there is a success message
:orange_circle: there is a warning
:red_circle: there is an error and the script halts 

#### in order
- defines some tool functions
- puts the first argument in `$file` and checks:
  - it was given as an argument :red_circle: 
  - it is a true file on the filesystem :red_circle:
- puts the rest of the arguments in an array and checks: 
  - the array is not empty :red_circle:
  - each user exists, only once, in the keyring :orange_circle:
  - shows the key if the user is valid in the keyring :green_circle:  
  - throws an error if any of them is invalid :red_circle: 
- encrypts the file with all the keys with `gpg` :large_blue_circle: 
- archives all the files in `$ARCHIVE` with `7z` :large_blue_circle:


scripts/gpg-share.sh
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env bash
+
+ARCHIVE="keys.7z"
```
atxr, line 3: Could you add the version, or maybe a date to avoid having several archives with the same name?

scripts/gpg-share.sh
```diff
@@ -0,0 +1,101 @@
...
+      log_info "encrypting '$file' for '$user'..."
+      gpg --verbose --recipient "$user" --encrypt --armor --output "$file.$user.asc" "$file"
+  done
+}
+
+
+# TODO: documentation
+archive () {
+  log_info "storing files in archive"
+  7zz a "$ARCHIVE" $(echo "${users[@]}" | tr ' ' '\n' | sed "s/\(.*\)/$file.\1.asc/")
```
atxr, line 94: 7zz? Same as 7z?
Could you add a dependency check for this?

atxr (changes requested):
Nice work! Thanks! :rocket:

Is it possible to add an option such as the script reads the users from a file?
That would be handy to avoid putting all the names of the team each time.

scripts/gpg-share.sh
amtoine: > Could you add the version, or maybe a date to avoid having several archives with the same name?
>
sure can :+1: 

i can propose the following
```diff
diff --git a/scripts/gpg-share.sh b/scripts/gpg-share.sh
index 49a424d..6a075fd 100755
--- a/scripts/gpg-share.sh
+++ b/scripts/gpg-share.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-ARCHIVE="keys.7z"
-
 
 # TODO: documentation
 log_error () {
@@ -27,6 +25,24 @@ log_info () {
 }
 
 
+# gives a default value to the VERSION file
+[ -z "$VERSION_FILE" ] && VERSION_FILE="package.json"
+
+# get the version
+VERSION=$(jq --raw-output .version "$VERSION_FILE" 2> /dev/null)
+[ -z "$VERSION" ] && {
+  log_error "$VERSION_FILE: No such file or directory"
+  exit 1
+}
+[ "$VERSION" = "null" ] && {
+  log_error "could not find 'version' in '$VERSION_FILE'"
+  exit 1
+}
+
+# build the archive name
+ARCHIVE="keys-${VERSION}.7z"
+
+
 # TODO: documentation
 check_file () {
   [ -z "$1" ] && {
```

amtoine: it simply uses `jq` to parse the `package.json` file

scripts/gpg-share.sh
amtoine: > 7zz? Same as 7z?
>
i do not know what `7z` is :thinking: 
i have
```bash
> pacman -Qo 7zz
/usr/bin/7zz is owned by 7-zip 22.01-2
```

> Could you add a dependency check for this?
>
yup, but what command should we use? `7zz`, `zip`, `7z`? :thinking:

scripts/gpg-share.sh
ctmbl: You could use `CURRENT_VERSION=$(npm pkg get version | sed 's/"//g')` that does exactly that job in my `bump.sh` script :wink:

scripts/gpg-share.sh
ctmbl: Hmmm
```bash
> pacman -Qo 7z
/usr/bin/7z is owned by p7zip 1:17.04-3
> pacman -Qo 7zz
error: No package owns 7zz
```

I think we should stick to `zip` because we don't need much compression and everyone has this one :+1:

ctmbl (changes requested):
I personally have nothing to say about the code in itself but after trying it I have 4 points:
 - as @atxr said it please provide a dependency check for `gpg` and the archive software
 - the script currently doesn't work well for user in two word, I tried with mine "Clement Mabileau" --> it interprets it as two different users and then creates two encrypted file. But that's normal looking at the code and I don't think that's important once you know it, you just have to specify only `./gpg-share.sh file Mabileau`!
 - currently the script isn't really clean because it leaves the user with multiple encrypted files at root without removing them after the archive is created (or `7zz` does it? I could find out because I don't have this one :confused:), consider either cleaning them or put them in `/tmp` I think!
 - you should add the `script` folder to the README repository structure I think!

scripts/gpg-share.sh
amtoine: perfect :ok_hand:

scripts/gpg-share.sh
amtoine: i do not have any `zip` command installed on my system :thinking:

scripts/gpg-share.sh
amtoine: should be added in fd81917

scripts/gpg-share.sh
amtoine: you tell me what command to use :relieved:

amtoine:
> * as @atxr said it please provide a dependency check for `gpg` and the archive software
>
the dependencies might change, e.g. `7zz`, but have a look at 6788388

> * the script currently doesn't work well for user in two word, I tried with mine "Clement Mabileau" --> it interprets it as two different users and then creates two encrypted file. But that's normal looking at the code and I don't think that's important once you know it, you just have to specify only `./gpg-share.sh file Mabileau`!
>
yep, that's how arrays work in `bash`... and i hate it :sob: 
i tried to fix this in b4eea59
when `USER_FILE` is set, the users are extracted from this file and not from `stdin`
```txt
# users.txt
Alexandre Tullot
amtoine
```
and
```bash
USER_FILE=users.txt ./scripts/gpg-share .env.production
```
should only try to use "`Alexandre Tullot`" and not "`Alexandre`" and "`Tullot`"
and the script should still work with `stdin` when `USER_FILE` is empty

> * currently the script isn't really clean because it leaves the user with multiple encrypted files at root without removing them after the archive is created (or `7zz` does it? I could find out because I don't have this one confused), consider either cleaning them or put them in `/tmp` I think!
>
i moved the files to `/tmp/` in 72582aa

> * you should add the `script` folder to the README repository structure I think!
>
added in 7c68495 :yum:

amtoine:
@atxr
> Is it possible to add an option such as the script reads the users from a file?
> That would be handy to avoid putting all the names of the team each time.
>
i've tried that in the previous comment and commits :wink:

scripts/gpg-share.sh
ctmbl: Seems good to me! @atxr ?

scripts/gpg-share.sh
ctmbl: Ok weird, what does `pacman -Qo zip` gives you? :open_mouth:

ctmbl:
> the dependencies might change, e.g. 7zz, but have a look at https://github.com/iScsc/iscsc.fr/commit/6788388b49dcf2f8fbe744bcbb1337f2cd3c6461

> i moved the files to /tmp/ in https://github.com/iScsc/iscsc.fr/commit/72582aa7224407baa553eea7fe031cd47a03441d

> added in https://github.com/iScsc/iscsc.fr/commit/7c68495fc5e695e5a2ff73e3e27731588da5dde6 yum

love these three tks!

However the script still doesn't work as expected, but that seems normal as you couldn't test it :confused: 
I found two issues actually.  
First running `./gpg-share.sh ../.env.example a2n-s` gives me this;
```
./gpg-share.sh: info: encrypting '../.env.example' for 'a2n-s'...
gpg: using pgp trust model
gpg: using subkey 99A3107C6A2A29D8 instead of primary key BFDE075AC06746A5
gpg: 99A3107C6A2A29D8: There is no assurance this key belongs to the named user

sub  cv25519/99A3107C6A2A29D8 2022-07-24 a2n-s <stevan.antoine@gmail.com>
 Primary key fingerprint: FBC3 7C3E 1887 74E9 FBD5  2FB9 BFDE 075A C067 46A5
      Subkey fingerprint: 565B 938E 3C91 902C 51FD  71BF 99A3 107C 6A2A 29D8

It is NOT certain that the key belongs to the person named
in the user ID.  If you *really* know what you are doing,
you may answer the next question with yes.

Use this key anyway? (y/N) y
gpg: reading from '../.env.example'
gpg: can't create '/tmp/../.env.example.a2n-s.asc': Permission denied
gpg: ../.env.example: encryption failed: Permission denied
```
So why does it says weird things about the the key not belonging to the named user?  
Second, at the end you'll understand that a simple `sed` isn't enough to properly create the key path :confused: **or the user should be aware that the file he is trying to encrypt must be at `$PWD`!** This problem of path of the file-to-encrypt can also lead to archiving `gpg-share.sh` instead of the encrypted files because of wrong path resolving...
```
./gpg-share.sh: info: storing files in archive
sed: -e expression #1, char 20: unknown option to `s'

7-Zip [64] 17.04 : Copyright (c) 1999-2021 Igor Pavlov : 2017-08-28
p7zip Version 17.04 (locale=en_US.UTF-8,Utf16=on,HugeFiles=on,64 bits,4 CPUs x64)

Scanning the drive:
1 file, 2566 bytes (3 KiB)

Creating archive: keys-0.1.0.7z

Items to compress: 1


Files read from disk: 1
Archive size: 1171 bytes (2 KiB)
Everything is Ok
./gpg-share.sh: info: content of the archive

7-Zip [64] 17.04 : Copyright (c) 1999-2021 Igor Pavlov : 2017-08-28
p7zip Version 17.04 (locale=en_US.UTF-8,Utf16=on,HugeFiles=on,64 bits,4 CPUs x64)

Scanning the drive for archives:
1 file, 1171 bytes (2 KiB)

Listing archive: keys-0.1.0.7z

--
Path = keys-0.1.0.7z
Type = 7z
Physical Size = 1171
Headers Size = 130
Method = LZMA2:12
Solid = -
Blocks = 1

   Date      Time    Attr         Size   Compressed  Name
------------------- ----- ------------ ------------  ------------------------
2022-11-14 00:06:15 ....A         2566         1041  gpg-share.sh
------------------- ----- ------------ ------------  ------------------------
2022-11-14 00:06:15               2566         1041  1 files
 ```
Or is it because of the weird `sed: -e expression #1, char 20: unknown option to 's'` ???

Also when using the `USER_FILE` variable the array creation seems to let a `\n` at the end of the user ids see:
```
./scripts/gpg-share.sh: success: key FBC37C3E188774E9FBD52FB9BFDE075AC06746A5 found for 'a2n-s'
./scripts/gpg-share.sh: success: key C4257605A0B2688C1CF2A04BB36E196507F03854 found for 'Clement Mabileau'
./scripts/gpg-share.sh: info: encrypting '.env.example' for 'a2n-s
'...
gpg: a2n-s
: skipped: No public key
gpg: .env.example: encryption failed: No public key
./scripts/gpg-share.sh: info: encrypting '.env.example' for 'Clement Mabileau
'...
gpg: Clement Mabileau
: skipped: No public key
gpg: .env.example: encryption failed: No public key
```
and 
```
Scanning the drive:

WARNING: No such file or directory
/tmp/.env.example.a2n-s.asc

WARNING: No such file or directory
/tmp/.env.example..asc


WARNING: No such file or directory
/tmp/.env.example.Clement.asc


WARNING: No such file or directory
/tmp/.env.example.Mabileau.asc


WARNING: No such file or directory
/tmp/.env.example..asc
```

Every of these bugs would have been found by you if you could run it on your computer :confused: , maybe you can create a VM or a container? or test it through ssh on the iscsc server?

File: scripts/gpg-share.sh (hunk 2 above)
Comment (amtoine) [reply scripts/gpg-share.sh]: > Ok weird, what does `pacman -Qo zip` gives you? open_mouth
>
```bash
> pacman -Qo zip
error: No package owns zip
```
it's just that `zip` is not installed by default on ArchLinux :+1:

amtoine:
@ctmbl 

i can test the script by applying
```diff
diff --git a/scripts/gpg-share.sh b/scripts/gpg-share.sh
index 3619dfd..300a4e4 100755
--- a/scripts/gpg-share.sh
+++ b/scripts/gpg-share.sh
@@ -26,7 +26,7 @@ log_info () {
 
 
 DEPENDENCIES=(
-  npm
+  # npm
   gpg
   7zz
 )
@@ -52,7 +52,8 @@ check_dependencies () {
 check_dependencies
 
 # get the version
-VERSION=$(npm pkg get version | sed 's/"//g')
+# VERSION=$(npm pkg get version | sed 's/"//g')
+VERSION="0.1.0"
 
 # build the archive name
 ARCHIVE="keys-${VERSION}.7z"
```
do not worry :wink: 

## `gpg` output
this is because one of the following i think:
- you never checked the fingerprint of your local `BFDE075AC06746A5` against mine, and did not `gpg --sign-key a2n-s`, so `gpg`  can not ensure the key is mine (see step 4 of [this tutorial](https://www.linuxbabe.com/security/gpg-guide-public-key-management))
- both `a2n-s` and `BFDE075AC06746A5` are deprecated and lost keys, so you should update your keyring with my new official personal public key => [7C5EE50BA27B86B7F9D5A7BA37AAE9B486CFF1AB](https://keys.openpgp.org/search?q=stevan.antoine@gmail.com) and then sign it of course

## the path
the script should be ran from the root of the repo
i did not test this from `./scripts/` and that's too much work for me for this PR...
the script does not exhibit this behaviour from the root, what about keeping that for a separate issue? :relieved: 

## the `USER_FILE` variable
yup, you're right, i'll have a look at that newline (super annoying :rage:)

amtoine:
if forgot a fix regarding the `USER_FILE` @ctmbl 
that's my fault, it should have worked directly in b4eea59 :yum: 

fix should be complete in 30f860c
arrays are terrible in `bash`, it's not the prettiest, with all these `sed` and `tr` manipulation and scope tricks, but i can not find a better solution for now...
i just do not get why there is a newline in the script when using `readarray` but not from a shell directly :shrug:

amtoine:
so, for now, please test the script only from the root :wink: 
it worked locally on my machine, with either `stdin` or a `USER_FILE` :+1:

scripts/gpg-share.sh
ctmbl: Okay that's a mess, after some discussion in private we'll go for `tar` as we hope that the one that most people will have

ctmbl: @amtoine

ctmbl:
@amtoine 

### `gpg` output

Thanks actually that was it, I'm not quite used to `gpg` right now that's true :sweat_smile: 

### the path 
> the script should be ran from the root of the repo
i did not test this from ./scripts/ and that's too much work for me for this PR...

I totally understand but we then should document this behavior :wink:, either as a **warning** at the beginning of the script (after checking `$PWD` or not) or in a **wiki**.

### the USER_FILE variable

Actually it now works at the encryption step:
```
./scripts/gpg-share.sh: info: encrypting '.env.example' for 'amtoine'...
gpg: using pgp trust model
gpg: using subkey 5CE93D5885AAB235 instead of primary key 37AAE9B486CFF1AB
gpg: This key probably belongs to the named user
gpg: reading from '.env.example'
gpg: writing to '/tmp/.env.example.amtoine.asc'
gpg: ECDH/AES256 encrypted for: "5CE93D5885AAB235 amtoine <stevan.antoine@gmail.com>"
./scripts/gpg-share.sh: info: encrypting '.env.example' for 'Clement Mabileau'...
gpg: using pgp trust model
gpg: using subkey F0443EE8B1A18913 instead of primary key B36E196507F03854
gpg: This key belongs to us
gpg: reading from '.env.example'
gpg: writing to '/tmp/.env.example.Clement Mabileau.asc'
gpg: ECDH/AES256 encrypted for: "F0443EE8B1A18913 Clement Mabileau <mabileau.clement@gmail.com>"
```
But still doesn't at the archiving step:
```
./scripts/gpg-share.sh: info: storing files in archive

7-Zip [64] 17.04 : Copyright (c) 1999-2021 Igor Pavlov : 2017-08-28
p7zip Version 17.04 (locale=en_US.UTF-8,Utf16=on,HugeFiles=on,64 bits,4 CPUs x64)

Scanning the drive:

WARNING: No such file or directory
/tmp/.env.example.Clement.asc


WARNING: No such file or directory
/tmp/.env.example.Mabileau.asc

1 file, 687 bytes (1 KiB)

Creating archive: keys-0.1.0.7z

Items to compress: 1
```

I know... these two words user ids are really annoying... sorry for that...

**Also**, could we pass the file as a command line argument instead of an environment variable? 
something like `-u amtoine` and `-U users.txt` to have a symetrical behavior while passing raw users or a file containing them
or is it too much work?

amtoine:
> ### `gpg` output
> Thanks actually that was it, I'm not quite used to `gpg` right now that's true sweat_smile
>
yep, this script assumes the keyring of the user is in good shape :thinking: 

### the path
> I totally understand but we then should document this behavior wink, either as a **warning** at the beginning of the script (after checking `$PWD` or not) or in a **wiki**.
> 
i've added a check in b069dd5
should terminate when not in the root of the repo :relieved: 

### the USER_FILE variable
> Actually it now works at the encryption step:
> 
cool :muscle: 
 
> But still doesn't at the archiving step:
> ...
> I know... these two words user ids are really annoying... sorry for that...
> 
i've changed the way i store the files in the archive:grey_exclamation: 
in f35fbd3, 8a6e589 and e8233a2, you can find the following:
- create a unique directory in `/tmp/` with `mktemp -u` and `mkdir`
- be explicite about the path to the encrypted files in the output of `encrypt`
- archive as follows
  - in a subshell, `cd` to the directory with encrypted files and archive the files by listing them with `find`
  - return automatically from the subshell
  - `cp` the archive back inside the root of the repo
 
that way
1. only the archive ends up in the repo
2.  the files are still available in `/tmp/`
3. there should not be issues with the names with `find`

> **Also**, could we pass the file as a command line argument instead of an environment variable? something like `-u amtoine` and `-U users.txt` to have a symetrical behavior while passing raw users or a file containing them or is it too much work?
>
i can think of that after the above changes are good to go :wink:

scripts/gpg-share.sh
amtoine: should be ok in https://github.com/iScsc/iscsc.fr/commit/f35fbd3ab5fb46daa8b7059a53b77be78eed13e1, https://github.com/iScsc/iscsc.fr/commit/8a6e5897524bb399430bdac1933d726bea390c1a and https://github.com/iScsc/iscsc.fr/commit/e8233a268f0cff16084d81abac7ec44c4661b092

amtoine:
mm, for the multi-word users....
i do not know it's a mess :sweat_smile: 

**what about not allowing them?**
it will be a lot easier and safer, to avoid files with spaces and so on...

atxr (approved):
You did an amazing work thank you for your contribution!
As @ctmbl said, it works fine for me with names of multiple words with this diff
```diff
- gpg --verbose --recipient "$user" --encrypt --armor --output "$DUMP_DIR/$file.$user.asc" "$file"
+ gpg --verbose --recipient "$user" --encrypt --armor --output $(echo "$DUMP_DIR/$file.$user.asc" | sed 's/\s\+/-/g') "$file"
```

Regarding the user option with `-u` and `-U` it would be great but that's up to you. This script solves the initial issue, so I'll approve this merge! :rocket:

amtoine:
@atxr @ctmbl 

glad it works :relieved: 

i've added the fix in the latest a04938a :+1: 
without the extra `sed` on `find` as you suggested :ok_hand: 

## flags -u and -U
> let me know what you want to do!
>
if that's ok with you both, i'd like to close this PR, as @atxr said it solves the original issue, and think of improvements in an issue i can open right after the merge :yum:

amtoine:
also, please solve the threads if they have been addressed :wink: 
otherwise, we won't be able to merge :eyes:

atxr:
> i've added the fix in the latest [a04938a](https://github.com/iScsc/iscsc.fr/commit/a04938a538187b6199108dcafdbb0067b031b2d4) +1 without the extra `sed` on `find` as you suggested ok_hand

Perfect :ok_hand: 

## flags -u and -U
> if that's ok with you both, i'd like to close this PR, as @atxr said it solves the original issue, and think of improvements in an issue i can open right after the merge :yum: 

Totally fine for me :wink: 
I resolved the remaining threads, we just need @ctmbl approval now.

ctmbl (commented):
It works perfectly tks a lot @amtoine for your amazing work!!
As for the improvements of course it can be address in other PRs

You'll just have to resolve some conflicts (I guess in the README) because I merged #27 before this one (ahah I'm evil :imp:)

amtoine:
@ctmbl, there it is :ok_hand:

ctmbl (approved):
Great! Ready to merge this one :rocket:

PR #48, opened 2022-11-20
FEATURE: add flag support to the `gpg-share` script

### LINKED ISSUE #37: improve the CLI of `gpg-share` with user, help and check flags
as introduced in #29, a new script has been added to the source to easily share the environment secrets with other developers using `pgp` keys and archives.

## the current script
the script is used as follows:
```bash
bash> ./scripts/gpg-share.sh <file-to-share> <user1> <user2> ...
```
or, when multi-words users have to be used or to save the list of people in a file
```bash
bash> USER_FILE="users.txt" ./scripts/gpg-share.sh <file-to-share>
```
with the following users
```txt
# users.txt
first user
another-one
last user
```
this is arguably not the best experience...

## idea of feature
during the review of #29, an idea has been introduced: **using flags**
- a `--user (-u)` flag to add one user from the standard input
- a `--users (-U)` flag to add additional users from a file

this will lead to the following usage:
```txt
# users.txt
additional user
```
```bash
bash> ./scripts/gpg-share.sh -u "first user" -u "user" -u "last one" -U users.txt
```
or 
```bash
bash> ./scripts/gpg-share.sh \
    -u "first user" \
    -u "user" \
    -u "last one" \
    -U users.txt
```
much better isn't it? :yum: 

we could also add `--help (-h)` to see the help of the script and why not `--check-keyring (-k)` to check to keyring of the user calling the script and make sure everything is ok :ok_hand:

amtoine:
This PR adds a bunch of flags to the `gpg-share.sh` script, to make its use both documented and easier.
This PR will close #37.

##  the changes
- **f55d62a..dfa0b31**: add flag support using the `getopt` command
:point_right: on dfa0b31, the script is equivalent to the one on 6d7a9f6, but with empty flag support
  - `git diff f55d62a^..dfa0b31` to see the diff
- **e83d952**: allow the user to pass the file and the list of users with `-f`, `-u` and `-U`
do not use `USER_FILE` anymore
  - `git show e83d952` to see the diff
- **da112f9..e30043d**: dedicated to the help and the usage
removes the `echo -n` lines thanks to the new better array system, see 9ec42b9 for details
  - `git diff da112f9^..e30043d:` to see the diff

## details about the usage
one can see the following
```bash
> ./scripts/gpg-share.sh -h
gpg-share:
     a script to share a secret file with others using pgp public keys.
     it should guide you when you make mistakes, e.g. providing a non-existing file.

Usage:
     gpg-share [-h] -f FILE {-u USER} {-U USER_FILE}

Options:
     -h/--help               shows this help.
     -f/--file               the secret file to share.
     -u/--user               a single user id in the keyring.
                               can be used multiple times.
                               supports multi-word ids inside quotes, e.g. "John Smith".
     -U/--users              the path to a file containing pgp ids.
                               should contain exactly one valid pgp id per line.
                               can be used multiple times.
```
the `-u` and `-U` flags can be used multiple times, and they are meant to be used as such
to avoid word splitting issues with `bash`, i chose to implement the flag parsing as follows:
- the user passes one user at the time with a separate `-u` each time. this allows to write `./scripts/gpg-share.sh -u "Alexandre Tullot"` and the id won't be split :star_struck: 
- to pass multiple users, simply write something like `./scripts/gpg-share.sh -u amtoine -u "Alexandre Tullot"`

this is the meaning of the `{ ... }` in the usage!
they mean the `-u` and `-U` flags can be used any number of times :+1: 

## what i did not do for now
- **`-u` and `-U` with multiple user ids**: this is too complex in `bash` according to me, i think this would require way to much work and might even be ambiguous as `bash` does not have typed variables, e.g. `-u amtoine "Alexandre Tullot" foo` might be treated as `"amtoine Alexandre Tullot foo"` :thinking:
in the end, i think the `-u SINGLE_USER_ID` syntax is as easy to write and far less ambiguous than the multi-id one
- **the `--check-keyring (-k)` option**: i do not know really what to check there :confused: 
- **exectute from `./scripts/`**: this was not the focus of this PR, so i did not change anything in that direction :+1: 
- **strong array checks**: i did not implement strong checks on the array of users, for instance, an id might appear multiple times in the array, e.g. with `-u amtoine -u amtoine` and the script will encrypt the secret file twice and overwrite previous encryptions for the `amtoine` user

### DIFF
--- a/scripts/gpg-share.sh
+++ b/scripts/gpg-share.sh
@@ -89,9 +89,8 @@ check_users() {
 
   # check if each of the "user" is in the keyring
   invalid=0
-  for _user in "${users[@]}";
+  for user in "${users[@]}";
   do
-      user=$(echo -n "$_user")
       nb_ids=$(gpg --quiet --list-keys "$user" 2> /dev/null | grep "^uid" | wc -l)
       case "$nb_ids" in
         0 ) log_warning "'$user' not found in the keyring..."; invalid=1 ;;
@@ -115,9 +114,8 @@ encrypt () {
   directory="$2"
   shift 2
   users=("$@")
-  for _user in "${users[@]}";
+  for user in "${users[@]}";
   do
-      user=$(echo -n "$_user")
       log_info "encrypting '$file' for '$user' inside '$directory'..."
       output=$(echo "$directory/$file.$user.asc" | sed 's/\s\+/-/g')
       gpg --verbose --recipient "$user" --encrypt --armor --output "$output" "$file"
--- a/scripts/gpg-share.sh
+++ b/scripts/gpg-share.sh
@@ -31,6 +31,8 @@ DEPENDENCIES=(
   tar
 )
 
+USAGE="gpg-share [-h]"
+
 
 # TODO: documentation
 check_dependencies () {
@@ -61,16 +63,6 @@ check_pwd () {
 }
 
 
-check_dependencies
-check_pwd
-
-# get the version
-VERSION=$(npm pkg get version | sed 's/"//g')
-
-# build the archive name
-ARCHIVE="keys-${VERSION}.tar"
-
-
 # TODO: documentation
 check_file () {
   [ -z "$1" ] && {
@@ -83,17 +75,6 @@ check_file () {
     exit 1
   }
 }
-
-
-file="$1"; shift 1
-check_file "$file"
-
-if [ -n "$USER_FILE" ]; then
-  check_file "$USER_FILE"
-  readarray users < "$USER_FILE"
-else
-  users=($@)
-fi
 
 
 # TODO: documentation
@@ -125,12 +106,6 @@ check_users() {
 }
 
 
-check_users
-
-DUMP_DIR=$(mktemp -u "/tmp/gpg-share-XXXXXX")
-mkdir -p "$DUMP_DIR"
-
-
 # TODO: documentation
 encrypt () {
   for _user in "${users[@]}";
@@ -156,5 +131,71 @@ archive () {
 }
 
 
-encrypt
-archive
+# parse the arguments.
+OPTIONS=$(getopt -o h --long help -n 'gpg-share' -- "$@")
+if [ $? != 0 ] ; then echo "Terminating..." >&2 ; exit 1 ; fi
+eval set -- "$OPTIONS"
+
+
+usage () {
+  echo "Usage: $USAGE"
+  echo "Type -h or --help for the full help."
+  exit 0
+}
+
+
+help () {
+  echo "gpg-share:"
+  echo "     TODO."
+  echo ""
+  echo "Usage:"
+  echo "     $USAGE"
+  echo ""
+  echo "Options:"
+  echo "     -h/--help               shows this help."
+  exit 0
+}
+
+
+main () {
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      -h | --help ) help ;;
+      -- ) shift; break ;;
+      * ) break ;;
+    esac
+  done
+
+  echo "ACTUAL STUFF COMING SOON!"
+  exit 0
+
+  check_dependencies
+  check_pwd
+
+  # get the version
+  VERSION=$(npm pkg get version | sed 's/"//g')
+
+  # build the archive name
+  ARCHIVE="keys-${VERSION}.tar"
+
+  file="$1"; shift 1
+  check_file "$file"
+
+  if [ -n "$USER_FILE" ]; then
+    check_file "$USER_FILE"
+    readarray users < "$USER_FILE"
+  else
+    users=($@)
+  fi
+
+  check_users
+
+  DUMP_DIR=$(mktemp -u "/tmp/gpg-share-XXXXXX")
+  mkdir -p "$DUMP_DIR"
+
+  encrypt
+  archive
+}
+
+
+main "$@"
--- a/scripts/gpg-share.sh
+++ b/scripts/gpg-share.sh
@@ -139,7 +139,7 @@ archive () {
 
 
 # parse the arguments.
-OPTIONS=$(getopt -o hu:U:f: --long help,user:,users:,file -n 'gpg-share' -- "$@")
+OPTIONS=$(getopt -o hu:U:f: --long help,user:,users:,file: -n 'gpg-share' -- "$@")
 if [ $? != 0 ] ; then echo "Terminating..." >&2 ; exit 1 ; fi
 eval set -- "$OPTIONS"
 
--- a/scripts/gpg-share.sh
+++ b/scripts/gpg-share.sh
@@ -66,7 +66,7 @@ check_pwd () {
 # TODO: documentation
 check_file () {
   [ -z "$1" ] && {
-    log_error "Please give a file to share as the first argument"
+    log_error 'Please give a file to share with the `-f` flag'
     usage
     exit 1
   }
@@ -82,7 +82,7 @@ check_file () {
 check_users() {
   users=("$@")
   [ "${#users[@]}" -eq 0 ] && {
-    log_error "Please give at least one gpg id after the first argument"
+    log_error 'Please give at least one gpg id with the `-u` and `-U` flags'
     usage
     exit 1
   }
--- a/scripts/gpg-share.sh
+++ b/scripts/gpg-share.sh
@@ -139,7 +139,7 @@ archive () {
 
 
 # parse the arguments.
-OPTIONS=$(getopt -o h --long help -n 'gpg-share' -- "$@")
+OPTIONS=$(getopt -o hu:U:f: --long help,user:,users:,file -n 'gpg-share' -- "$@")
 if [ $? != 0 ] ; then echo "Terminating..." >&2 ; exit 1 ; fi
 eval set -- "$OPTIONS"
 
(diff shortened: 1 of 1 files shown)


scripts/gpg-share.sh
```diff
@@ -85,29 +78,19 @@ check_file () {
...
-else
-  users=($@)
-fi
-
-
 # TODO: documentation
 check_users() {
+  users=("$@")
   [ "${#users[@]}" -eq 0 ] && {
     log_error "Please give at least one gpg id after the first argument"
```
ctmbl, line 85: Now that flags has been introduced you should change this to :
```diff
-    log_error "Please give at least one gpg id after the first argument"
+    log_error 'Please give at least one gpg id with `-u` or `-U` flags'
```

scripts/gpg-share.sh
```diff
@@ -61,20 +63,11 @@ check_pwd () {
...
-VERSION=$(npm pkg get version | sed 's/"//g')
-
-# build the archive name
-ARCHIVE="keys-${VERSION}.tar"
-
-
 # TODO: documentation
 check_file () {
   [ -z "$1" ] && {
     log_error "Please give a file to share as the first argument"
```
ctmbl, line 69: *same as the other error msg*
Now that flags has been introduced you should change this to :
```diff
-    log_error "Please give a file to share as the first argument"
+    log_error 'Please give a file to share with `-f` flag'

File: scripts/gpg-share.sh [hunk 3]
```diff
@@ -125,36 +108,105 @@ check_users() {
...
+    cd "$directory"
+    tar cvf "$name" $(find . -type f)
   )
-  cp "$DUMP_DIR/$ARCHIVE" .
+  cp "$directory/$name" .
+}
+
+
+# parse the arguments.
+OPTIONS=$(getopt -o hu:U:f: --long help,user:,users:,file -n 'gpg-share' -- "$@")
```
ctmbl, line 142: should be:
```diff
-OPTIONS=$(getopt -o hu:U:f: --long help,user:,users:,file -n 'gpg-share' -- "$@")
+OPTIONS=$(getopt -o hu:U:f: --long help,user:,users:,file: -n 'gpg-share' -- "$@")
```
I think!

scripts/gpg-share.sh
```diff
@@ -125,36 +108,105 @@ check_users() {
...
+  exit 0
+}
+
+
+main () {
+  users=()
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      -h | --help ) help ;;
+      -f | --file ) file="$2"; shift 2;;
```
ctmbl, line 180: maybe add a check to forbid multiple use of `-f`? (which would end up not stacking but overwriting previous arg)

scripts/gpg-share.sh
```diff
@@ -125,36 +108,105 @@ check_users() {
...
+      -u | --user ) users+=("$2"); shift 2;;
+      -U | --users ) 
+        check_file "$2"
+        readarray new_users < "$2"
+        shift 2
+        for new_user in "${new_users[@]}"; do
+          users+=("$(echo -n "$new_user")")
+        done;;
+      -- ) shift; break ;;
+      * ) break ;;
```
ctmbl, line 190: maybe add a "unknown arg"? or is it already handled by `getopt`?

scripts/gpg-share.sh
```diff
@@ -125,36 +108,105 @@ check_users() {
...
+    esac
+  done
+
+  check_dependencies
+  check_pwd
+
+  check_file "$file"
```
ctmbl, line 197: Maybe we should do this in arg parsing as for the `-U` option!

ctmbl (changes requested):
Tks for this great update :wink:! Still some questions and changes missing but that seems good!

>-u and -U with multiple user ids: this is too complex in bash according to me, i think this would require way to much work and might even be ambiguous as bash does not have typed variables, e.g. -u amtoine "Alexandre Tullot" foo might be treated as "amtoine Alexandre Tullot foo" thinking
in the end, i think the -u SINGLE_USER_ID syntax is as easy to write and far less ambiguous than the multi-id one

I completely agree!

>the --check-keyring (-k) option: i do not know really what to check there confused

I guess you could:
 - ask for a `-u` and `-U` as well and check that their keys are in the user keyring
 - check that these keys are signed up

scripts/gpg-share.sh
```diff
@@ -85,29 +78,19 @@ check_file () {
...
     exit 1
   }
 
   # check if each of the "user" is in the keyring
   invalid=0
-  for _user in "${users[@]}";
+  for user in "${users[@]}";
   do
-      user=$(echo -n "$_user")
       nb_ids=$(gpg --quiet --list-keys "$user" 2> /dev/null | grep "^uid" | wc -l)
```
ctmbl, line 94: Actually I just discovered that you should use `--with-colons` here
> Avoid using the output of this command in scripts or other programs as it is likely to change as GnuPG changes. See --with-colons for a machine-parseable key listing command that is appropriate for use in scripts and other programs. 

From [gpg man page](https://manpages.org/gpg)

ctmbl (commented):
Maybe something for another PR but seems worth saying!

scripts/gpg-share.sh
amtoine: not sure i got what you propose :thinking: 

- do you want to **throw an error when `-f` is used multiple times**?
- do you want to allow the use of multiple `-f`s and do as with the users and **build an array of files to encrypt and share**?

scripts/gpg-share.sh
amtoine: yup it is :yum: 

see
```bash
> ./scripts/gpg-share.sh --djakldjlka
gpg-share: unrecognized option '--djakldjlka'
Terminating...
```

scripts/gpg-share.sh
amtoine: i am used to parsing all the arguments and then running the actual script :thinking: 

i think it's better to separate the two to avoid subtle bugs :confused:

scripts/gpg-share.sh
amtoine: should be fixed in ca775a9

scripts/gpg-share.sh
amtoine: should be fixed in 29d6f0f

scripts/gpg-share.sh
amtoine: actually, this is super interesting :open_mouth: :smirk: 

however, i simply want to count the numbers of keys with the same user name as the input :thinking: 
and as each key entry has a unique "_uid_" at the beginning of the line, the other one being "_pub_", "_sig_", "_sig 3_" and "_sub_"

in the end, i simply do not see where this will improve the count and make the script "safe" :yum:

scripts/gpg-share.sh
ctmbl: I would like to **forbid** it so throwing an error seems good :+1:

scripts/gpg-share.sh
ctmbl: perfect!

scripts/gpg-share.sh
ctmbl: I get your point :yum: just wanted to have symmetrical code between `-U` handling and `-f` (even if I get that you're **forced** to do it the `switch case` for `-U`) 
Feel free to resolve this if you want to stand your ground :+1:

scripts/gpg-share.sh
ctmbl: That's right actually it won't (at least should not) make the script safer!

But I wanted to point this out to share my discovery **and** because even if it doesn't improve the script: it's still good practice :yum: 
(example: someone else wants to modify the script and doesn't know this flag, if you already use it he is more likely to use it too! making the script better :wink: )

atxr (approved):
I tested this new feature and it is great!
@ctmbl already did a good review job, I don't have anything to add in the scope of this PR
Nevertheless, I have some improvement ideas for this script, but I'll open another issue.
Thanks! :rocket:

scripts/gpg-share.sh
amtoine: so we'll encrypt a single file then :wink:

scripts/gpg-share.sh
amtoine: > I get your point yum just wanted to have symmetrical code between `-U` handling and `-f` (even if I get that you're **forced** to do it the `switch case` for `-U`)
>
yup, i think it's far better and easier to build the array of users directly in the parsing.
this is a very low overhead to the parsing of the arguments and imo the goal of parsing the arguments here is to 
- have the filename
- have the full array of users

> Feel free to resolve this if you want to stand your ground +1
>
i'll let the parsing as it is then :relieved:

scripts/gpg-share.sh
amtoine: i love what you said.
so i will use `gpg --with-colons` for this :muscle:

scripts/gpg-share.sh
amtoine: fix available in 86de353

scripts/gpg-share.sh
amtoine: should be good to go in 67a797e :yum: 

i've still been able to encrypt the `README` into an archive, share it to myself (lol), un`tar` it, decrypt  it with `gpg` and read the result :ok_hand: 
and the script checks still look to work :+1:

ctmbl (approved):
Perfect! I've nothing to add, the feature is really cool according to me! 

I think we're ready to merge, I however suggest that we wait some time for @vSt11 review/approval (in an educational will), but @amtoine feel free to merge this PR if @vSt11 doesn't answer in the coming days :wink:
Thanks for your work once again an amazing job :rocket:

amtoine:
> Thanks for your work once again an amazing job rocket
>
glad you like it :relieved: 

> I think we're ready to merge, I however suggest that we wait some time for @vSt11 review/approval (in an educational will), but @amtoine feel free to merge this PR if @vSt11 doesn't answer in the coming days wink
> 
no no, there's no rush, this PR can wait for @vSt11 :yum:

PR #49, opened 2022-11-22
README update - node, nodemon and npm versions

vSt11:
Everything is in the title, resolving a minor issue Tell me if it looks good to you ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ npm install
 npm run dev
 ```
 
-> You will need `nodemon` to run the backend. Use `npm install -g nodemon` to install it.
+> You will need `nodemon` to run the backend. Use `npm install -g nodemon` to install it. Make sure you're supporting at least 2.0.20 with `nodemon --version`. Nodemon has been tested working fine with node 19.
 
 #### Frontend
 
@@ -91,6 +91,8 @@ cd frontend
 npm install
 npm run start
 ```
+Make sure your're using at least version 8.19.2 by checking `npm --version`, and update if needed with `npm update`.
+ 
 
 ### Production mode
 


ctmbl (approved):
This seems fine to me :+1: but of course we should wait for @atxr approval because he is the most experienced of us when it comes to `node` and `npm`

amtoine:
one thing i do not understand about this PR :thinking: 

**the website has become quite complex and is supposed to be run in a `docker` container only, right?**
or am i wrong on that point? :confused: 

let's assume i'm right and that the application is not supposed to be run manually from the command line outside of the `docker-compose` container
then the versions of `node` and `npm` are the ones specified inside the `docker` image, right? e.g. `FROM node:18` will force the base image to be `node:18` with something like `ubuntu` or `alpine` running, a bunch of tools and `node` and `npm` with versions coming from the base image `node:18` but not from the host system
with all that, and that is my current understanding of how both the website and `docker` are supposed to work :+1:, i don't think we should tell the user to fix their versions of `node` and `npm`, because the locally installed `node` and `npm` won't ever be used inside the containers, only the ones from `node:18` will!

in the end, i think we should enforce the versions from inside the `docker` images:
- `docker-compose.yml`
- `frontend/Dockerfile`
- `backend/Dockerfile`

but not ask the user to change their `node` and `npm` in the `README` :+1: 

again, correct me if i'm wrong :relieved: 
the changes here might be perfect if what i'm saying is not how things work :wink: 
but for now, that looks more confusing to me :confused:

amtoine:
then i propose something to make this PR much stronger:
- write two scripts, to run the backend and the frontend respectively
- update the `README` to just call the scripts instead of a collection of single commands
- inside the scripts, enforce the versions of `node` and `npm` are correct with tools like `semver` for instance

## precision
this can be moved to another issue if you feel it's too much work :wink: 

just ping me with your thoughts and i'll move that comment to some other place if it's out of the scope of this PR :relieved:

README.md
```diff
@@ -80,7 +80,7 @@ npm install
 npm run dev
 ```
 
-> You will need `nodemon` to run the backend. Use `npm install -g nodemon` to install it.
+> You will need `nodemon` to run the backend. Use `npm install -g nodemon` to install it. Make sure you're supporting at least 2.0.20 with `nodemon --version`. Nodemon has been tested working fine with node 19.
```
atxr, line 83: I actually use nodemon 2.0.19 :eyes: 
But that's a minor update so it shouldn't be a big issue

atxr (approved):
I commented the nodemon version because I use 2.0.19, but I'll approve this PR because it's a trivial detail.

atxr:
Aha, yes as you highlighted, the development mode doesn't use docker at all for now.

> then i propose something to make this PR much stronger:
> 
>     * write two scripts, to run the backend and the frontend respectively
> 
>     * update the `README` to just call the scripts instead of a collection of single commands
> 
>     * inside the scripts, enforce the versions of `node` and `npm` are correct with tools like `semver` for instance
 
Those are great ideas to improve the dev workflow. However, it's totally out of the scope of this PR, because you need some knowledge about npm, and the development pipeline.
Moreover, with @ctmbl, we plan to move the dev mode to docker as well, so I don't think we need to implement these points.

amtoine:
> Aha, yes as you highlighted, the development mode doesn't use docker at all for now.
>
yup my bad :relieved: 
 
> Those are great ideas to improve the dev workflow. However, it's totally out of the scope of this PR, because you need some knowledge about npm, and the development pipeline.
>
got it :+1: 

> Moreover, with @ctmbl, we plan to move the dev mode to docker as well, so I don't think we need to implement these points.
>
then i think the `README` will have to be updated to mitigate the confusion :yum:

README.md
ctmbl: @vSt11

vSt11:
I don't have any writing access, so someone that does can merge this branch now (I think)

amtoine:
@atxr @vSt11 @ctmbl 
should that PR be closed or merged? :confused: :sweat_smile:

vSt11:
I don't have the right to reopen it, but it should be merged 😆  I wondered if this should be closed before merging but i got my answer now

ctmbl:
Well this one was supposed to be merged that's right :laughing:

ctmbl (approved):

ctmbl:
@atxr I think @vSt11 wants to see its work merged, you just need to re-approve, no new commits have been added just the reopen of the PR :+1:

atxr (approved):
Okay no problem! :racehorse:

PR #61, opened 2022-12-11
Fix backend header set twice

### LINKED ISSUE #60: Error HTTP Headers already set
Recently, @ctmbl noticed in the backend logs this error:
```
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at new NodeError (node:internal/errors:393:5)
    at ServerResponse.setHeader (node:_http_outgoing:644:11)
    at ServerResponse.header (/backend/node_modules/express/lib/response.js:794:10)
    at ServerResponse.send (/backend/node_modules/express/lib/response.js:174:12)
    at ServerResponse.json (/backend/node_modules/express/lib/response.js:278:15)
    at getByAuthor (/backend/controllers/articleController.js:23:21)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'ERR_HTTP_HEADERS_SENT'
}
```

atxr:
* Return if error to avoid setting the headers twice
Close #60

### DIFF
--- a/backend/controllers/articleController.js
+++ b/backend/controllers/articleController.js
@@ -14,6 +14,7 @@ const getByAuthor = async (req, res) => {
   const author = req.params.author ?? req.user?.usermame
   if (!author) {
     res.status(404).json({error: "You must login or provide an author" });
+    return
   }
 
   try {


ctmbl (approved):
Didn't test it but seems quite logic. 
If you tested it right we can merge it :+1:

amtoine (commented):
as @ctmbl said, if you tested it, i'm fine with this :relieved: 

### just a question, and i'll approve
what is the new `return` supposed to do? :thinking: 
it only prevents the bug in #60 while keeping the rest of the website intact? :open_mouth:

atxr:
The function `res.status` or `res.sendStatus` can be used only once in a route. It basically sets the response headers.
Here, without this `return`, we set the headers to `404` once and then still execute the remaining code, which will set the headers again at the end.
So this return allows to quit the function when no `author` is provided.

amtoine (approved):
> The function `res.status` or `res.sendStatus` can be used only once in a route. It basically sets the response headers.
Here, without this `return`, this function might set the headers to `404` once and then still execute the remaining code, which will set the headers again at the end.
So this return allows to quit the function when no `author` is provided. 
>
ok that makes total sense, thanks for the precision :relieved: 

then it's great imo :+1:

amtoine:
again, waiting a bit for @gbrivady :yum: 
@atxr, if you feel time has passed enough, just merge this :wink:

atxr:
Shouldn't we wait for #66 before merging this one?

amtoine:
> Shouldn't we wait for #66 before merging this one?
>
imo, this is a bug fix so, even if it was not planned in the [`0.1.1`](https://github.com/iScsc/iscsc.fr/milestone/3), it still is a bug fix and thus could go inside this patch version ... or after

in the end, that does not look like a big deal at all :wink:

gbrivady (approved):
Seems fine to me

Issue #44, opened 2022-11-18
Blog posts can't be shared by their URL


ctmbl:
#### Problem
The website has been deployed, and I was trying to visit a blog post from `amtoine` (the first actually) by clicking on an URL

#### Step to reproduce:
 - Try to access to https://iscsc.fr/blog/637669b524bc362fff7a2b23
 - go to https://iscsc.fr and choose the blog post from `amtoine - 2022-11-17T17:04:53.781Z`
 - the URL are the same but accessing with the URL seems to fail

gbrivady:
Probably related to #15

gbrivady:
Adding my findings after some investigation, but basically seems like **there is an issue with either the `Article.js` page; or the routing.**

If you access a blog page directly from its url, and look at the console, you have the following log: [iscsc.fr-1670535866067.log](https://github.com/iScsc/iscsc.fr/files/10189672/iscsc.fr-1670535866067.log), with an error, from which you can deduce what I commented, on the states of everything in Article:

```js
const Article = () => {
  let { id } = useParams();
  const { articles } = useArticlesContext(); //After this line, articles is null
  const article = articles.filter((a) => a._id === id)[0]; //So this line generates the error
  return <ArticleView article={article} />;
};
```

Except that if you look at the log when accessing a blog post from the main page, `articles` is not null (you can check by using `console.log(articles)` if you wish to reproduce), the log is clean.

Furthermore, if you go on the main page, open an article, then refresh the page, you will find the same empty page, same error. And this error is also there for #15.

So I can deduce the following:

- #15 and this issue are the same issue
- **It seems like articles are fetched from the  database only if you first go through the main page**

ctmbl:
@gbrivady 
> So I can deduce the following:
Opening an article in a new tab or window displays a blank page #15 and this issue are the same issue

Actually I came to the the same conclusion while performing my own tests :confused: 
However check https://iscsc.fr/api/articles/638fdec624bc362fff7a2d18 it doesn't seem to have a problem with the routes in the backend for sure. I'd probably blame the `ArticleContext.js` file because at some point it "creates a context" (whatever that means in React) and initialize what will be returnes to null. But I can't find what's missing

ctmbl:
Ol I think I got it
In `Article.js` the code get `articles` from the context.
However if we access the post by URL the context is initialize at null: `dispatch SET` has never been sent and articles have not been fetched from teh API, when accessing from /blog, this `Blog.js` code is executed:
[...]

PR #63, opened 2022-12-15
Fix empty context array when directly browsing to an URL

### LINKED ISSUE #15: Opening an article in a new tab or window displays a blank page
Title is self explanatory. Right-click -> "Open in a new tab/window" displays a blank page instead of the article.
Issue happens on both Chrome Version 105/106 and Firefox Version 105 on Windows 10.

**Details:**
- OS: Windows 10
- npm version: 8.19.1
- node version: 18.9.1

gbrivady:
Fixes #15 and #44 

Basically implemented the fix suggested by @atxr in #44, with proper async'ing of functions and stuff

### DIFF
--- a/frontend/src/context/ArticleContext.js
+++ b/frontend/src/context/ArticleContext.js
@@ -10,9 +10,10 @@ export const articlesReducer = (state, action) => {
       };
 
     case "ADD":
-      return {
-        articles: [action.payload, ...state.articles],
-      };
+      if (state.articles == null) {
+        return {articles: [action.payload], }; //cannot iterate over null
+      }
+      return {articles: [action.payload, ...state.articles], };
 
     case "DELETE":
       return {
--- a/frontend/src/context/ArticleContext.js
+++ b/frontend/src/context/ArticleContext.js
@@ -10,7 +10,7 @@ export const articlesReducer = (state, action) => {
       };
 
     case "ADD":
-      if (state.articles == null) {
+      if (state.articles === null) {
         return {articles: [action.payload], }; //cannot iterate over null
       }
       return {articles: [action.payload, ...state.articles], };
--- a/frontend/src/pages/blog/Article.js
+++ b/frontend/src/pages/blog/Article.js
@@ -4,9 +4,28 @@ import { useArticlesContext } from "../../hooks/useArticlesContext";
 
 const Article = () => {
   let { id } = useParams();
-  const { articles } = useArticlesContext();
-  const article = articles.filter((a) => a._id === id)[0];
-  return <ArticleView article={article} />;
+  const { articles, dispatch } = useArticlesContext();
+
+  const fetchArticle = async () => {
+      const response = await fetch(`/api/articles/${id}`);  // fetch the article (only this one) from the database
+      if (response.ok) {
+        const article = await response.json();
+        dispatch({ type: "ADD", payload: article });       // if it exists, add it to the ArticlesContext
+        return article;
+      }
+      else {
+        return null;
+      } 
+  };
+  
+  try {
+    const article = articles.filter((a) => a._id === id)[0];
+    return <ArticleView article={article} />;
+  } catch (error) { //if articles are null, articles.filter produces an error
+    const article = fetchArticle();
+    return <ArticleView article={article} />;
+  }
+
 };
 
 export default Article;
--- a/frontend/src/pages/blog/Article.js
+++ b/frontend/src/pages/blog/Article.js
@@ -18,13 +18,14 @@ const Article = () => {
       } 
   };
   
+  let article;
   try {
-    const article = articles.filter((a) => a._id === id)[0];
-    return <ArticleView article={article} />;
+    article = articles.find((a) => a._id === id);
   } catch (error) { //if articles are null, articles.filter produces an error
-    const article = fetchArticle();
+    article = fetchArticle();
     return <ArticleView article={article} />;
   }
+  return <ArticleView article={article} />;
 
 };
 


amtoine:
i added #44 to the links, you forgot it :wink:

ctmbl:
Thanks for this fix it was pretty annoying! I'm quite busy right now but I'll look into it as soon as I can!

frontend/src/pages/blog/Article.js
```diff
@@ -4,9 +4,28 @@ import { useArticlesContext } from "../../hooks/useArticlesContext";
...
+        const article = await response.json();
+        dispatch({ type: "ADD", payload: article });       // if it exists, add it to the ArticlesContext
+        return article;
+      }
+      else {
+        return null;
+      } 
+  };
+  
+  try {
```
atxr, line 21: Just some minor comments here:
- I would use `let` instead of `const`
- You should move this return outside of the try/catch
- I found a cleaner way to do `filter(...)[0]`

```js
let article;
try {
    article = articles.find((a) => a._id === id);
} catch (error) {  //if articles are null, articles.filter produces an error
    article = fetchArticle();
}

return <ArticleView article={article} />;
```

atxr (changes requested):
Thanks for the implementation!
I added few suggestions, and I will test once they will be solved.

frontend/src/context/ArticleContext.js
```diff
@@ -10,9 +10,10 @@ export const articlesReducer = (state, action) => {
       };
 
     case "ADD":
-      return {
-        articles: [action.payload, ...state.articles],
-      };
+      if (state.articles == null) {
+        return {articles: [action.payload], }; //cannot iterate over null
+      }
+      return {articles: [action.payload, ...state.articles], };
```
ctmbl, line 16: Perhaps you could do something like
```js
return state.articles ? {articles: [action.payload, ...state.articles], } : {articles: [action.payload], };
```
or something using the [`??` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)

Maybe it would sacrifice too much the code clearness :confused: 

At least we could use a `===` instead of  `==`
But I'm not quite sure so once again @atxr :smiling_face_with_tear: ?

ctmbl: My bad I actually didn't quite understand the real problem, see my other thread on `dispatch` call!

frontend/src/pages/blog/Article.js
```diff
@@ -4,9 +4,28 @@ import { useArticlesContext } from "../../hooks/useArticlesContext";
...
-  const { articles } = useArticlesContext();
-  const article = articles.filter((a) => a._id === id)[0];
-  return <ArticleView article={article} />;
+  const { articles, dispatch } = useArticlesContext();
+
+  const fetchArticle = async () => {
+      const response = await fetch(`/api/articles/${id}`);  // fetch the article (only this one) from the database
+      if (response.ok) {
+        const article = await response.json();
+        dispatch({ type: "ADD", payload: article });       // if it exists, add it to the ArticlesContext
```
ctmbl, line 13: I think you should do
```suggestion
        dispatch({ type: "SET", payload: [article] }); // if it exists, add it to the ArticlesContext
```
as `SET` is defined (and used in `frontend/src/pages/blog/Blog.js`) it is the proper way to do what you are trying to do!
And it also remove the need of redefining `ADD` as you did :wink: 
@atxr

ctmbl (changes requested):
Some changes but overall it simply solves a really annoying issue so thanks a lot!

frontend/src/pages/blog/Article.js
```diff
@@ -4,9 +4,28 @@ import { useArticlesContext } from "../../hooks/useArticlesContext";
...
+      }
+      else {
+        return null;
+      } 
+  };
+  
+  try {
+    const article = articles.filter((a) => a._id === id)[0];
+    return <ArticleView article={article} />;
+  } catch (error) { //if articles are null, articles.filter produces an error
```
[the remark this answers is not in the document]
atxr: What kind of error would you like to filter?

frontend/src/pages/blog/Article.js
atxr: No I don't think this is a good idea.
Actually if your context is not empty it will overwrite it.
Take the following example:
-> You browse an article
-> From this article, you browse another article
-> You finally go back to the first article
In this example, if you `SET`, the first article will be overwritten when browsing the second one. So the app will fetch again the first article in the third step.
I hope it makes sense :)

frontend/src/context/ArticleContext.js
atxr: A real *Never Nester* would opt for
```js
state.articles ??= [] 
return {articles: [action.payload, ...state.articles], };
```

frontend/src/pages/blog/Article.js
gbrivady: @atxr answered in my stead, if you use `SET` for some reason but there are already articles loaded in, you will overwrite everything - so definetly not a good idea

frontend/src/pages/blog/Article.js
gbrivady: Let's say that my "limited `JS` knowledge created this infamy 😅 fixed in 0d870937c5d299647f9a40d996205c7e7020bdf0

frontend/src/context/ArticleContext.js
gbrivady: Not exactly enlightened in `JS`, so I was unaware of all the equality stuff, but I do now 👍 fixed in 3b178585584f276ceabc0c56b954f8a59510594d

Not going to used any `??` shenanigan tho because I think it's not necessary, the nesting is not excessive I believe, so let's keep it clear 😉

frontend/src/pages/blog/Article.js
atxr: Perfect!

frontend/src/context/ArticleContext.js
atxr: Perfect good for me!

atxr: I'll let @ctmbl resolve

atxr (approved):
🚀

frontend/src/pages/blog/Article.js
ctmbl: @atxr 
Ok you're right I didn't see things this way, but then, what's the whole point of *having* a `SET`?

frontend/src/context/ArticleContext.js
ctmbl: > A real Never Nester would opt for

Ow god that was exactly what I was looking for but didn't manage to find it :heart_eyes:

frontend/src/pages/blog/Article.js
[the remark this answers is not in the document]
ctmbl: I don't know the names of errors that can happen but you never now what can trigger an error so better be safe than sorry right? 
In `python` even if I can't *imagine* what kind of Exception could occur I would still filter it as a matter of principle :+1:

frontend/src/pages/blog/Article.js
ctmbl: Actually I'm wondering: why don't we need to pass `id` as parameter of `fetchArticle`?

gbrivady: `id`is defined within the scop of the `Article` class (see line 1) so it already sees it

ctmbl: I get it but it feels weird I guess!

frontend/src/pages/blog/Article.js
atxr: `SET` is used [here](https://github.com/iScsc/iscsc.fr/blob/main/frontend/src/pages/blog/Blog.js#L15)
It's used when you load the `/` route to fetch all the article.

ctmbl: Yes I already saw it but now that `ADD` has been patched what's the difference between using `ADD` when empty and using `SET` (apart from the list), does it not seem confusing to you ?

atxr: `ADD` adds only one article when `SET` can add multiple at the same time

frontend/src/pages/blog/Article.js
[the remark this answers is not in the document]
atxr: You are right, but here, regardless the error, we should fetch the article again. 
We don't really need to switch on different type of exception, because in every case, we should fetch.

atxr (approved):

frontend/src/pages/blog/Article.js
ctmbl: Ok, I already got this point but while thinking about yesterday I finally understand why we need both, my problem was a little pointy but I get it!

frontend/src/pages/blog/Article.js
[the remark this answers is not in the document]
ctmbl: Ok tks for the explanation, I'm ok with it so let's go

ctmbl (approved):
Nice fix, and sorry for the delay :sweat_smile: let's merge it :rocket:

PR #70, opened 2023-03-08
Remove intermediary commit in `bump.sh`

ctmbl:
This PR simply implement one of the decision that was taken after last release: 
 - the automatic bump script shouldn't commit twice and shouldn't tag
It also corrects minor problems.
Once the PR is merge, the documentation will be updated accordingly

Another bigger PR is coming that should implement the other decision:
 -  add `patch` `minor` and `major` subcommands to the script to easily bump version *Ex: currently (at 0.1.1) `./bump.sh patch` would be equivalent as `./bump.sh 0.1.2`*
(and we'll probably add better arguments handling)

### DIFF
--- a/scripts/bump.sh
+++ b/scripts/bump.sh
@@ -150,5 +150,5 @@ log_info 'Pushing branch and new version tag'
 log_warning "pushing to \`${ISCSC_REMOTE}\` please type your passphrase/password if required:"
 [ -z "$DRY_RUN" ] && { git push ${ISCSC_REMOTE} ${BUMP_BRANCH} v${CURRENT_VERSION} || log_error "push failed, you can push with \`git push ${ISCSC_REMOTE} ${BUMP_BRANCH} v${CURRENT_VERSION}\`"; }
 
-log_error '`npm install` has been run during the bump, you MUST review the changes during PR review to ensure package.json and package-lock.json where compatible!!!'
+log_hint '`npm install` has been run during the bump, you MUST review the changes during PR review to ensure package.json and package-lock.json where compatible!!!'
 
--- a/scripts/bump.sh
+++ b/scripts/bump.sh
@@ -122,8 +122,9 @@ log_info "'${NEW_VERSION}'>'${CURRENT_VERSION}', '${NEW_VERSION}' is accepted as
 # --------------------------------- Git setup ----------------------------------
 
 # ...and checkout on main to create a version bump branch
+# (working dir is empty check passed)
 log_info "Checkout on ${ISCSC_REMOTE}/main"
-[ -z "$DRY_RUN" ] && { git checkout ${ISCSC_REMOTE}/main || exit 1; }
+[ -z "$DRY_RUN" ] && git checkout ${ISCSC_REMOTE}/main
 log_info "switching to ${BUMP_BRANCH}"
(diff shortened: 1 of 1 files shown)


scripts/bump.sh
```diff
@@ -141,17 +142,16 @@ log_info 'Bumping `backend`'
...
 
-log_info 'Commiting `frontend` and `backend` bump'
-[ -z "$DRY_RUN" ] && { git commit -m "Bump frontend and backend versions to ${NEW_VERSION}" || exit 1; }
-
 # ----------------------------- Bump root and push -----------------------------
 
 log_info 'Bumping `root`'
-[ -z "$DRY_RUN" ] && { npm version "${NEW_VERSION}" -m "Bump to version %s" || exit 1; }
+[ -z "$DRY_RUN" ] && { npm version "${NEW_VERSION}" --no-git-tag-version || exit 1; }
+[ -z "$DRY_RUN" ] && { git commit -m "Bump to version ${NEW_VERSION}" || exit 1; }
```
amtoine, line 149: is this the commit?

'cause it only looks like it moved, not it was removed :thinking:

amtoine (commented):
a question about the commit we talk about.

scripts/bump.sh
ctmbl: No it didn't just moved! The first commit has been removed [here](https://github.com/iScsc/iscsc.fr/pull/70/commits/a39a1034d56926c17bbcea14f74c92a017fd61be) and the 2nd was automatically made by `npm version` --> I added `--no-git-tag-version` to every `npm version` call, then the only commit now is the `git commit` at line 149

amtoine: aaah

`npm version ... -m ...` does a `git` commit?
that would explain the "disappearing commit" :+1:

ctmbl: Exactly!

atxr (approved):
Looks legit! 🤖

scripts/bump.sh
```diff
@@ -122,8 +122,9 @@ log_info "'${NEW_VERSION}'>'${CURRENT_VERSION}', '${NEW_VERSION}' is accepted as
 # --------------------------------- Git setup ----------------------------------
 
 # ...and checkout on main to create a version bump branch
+# (working dir is empty check passed)
 log_info "Checkout on ${ISCSC_REMOTE}/main"
-[ -z "$DRY_RUN" ] && git checkout ${ISCSC_REMOTE}/main || exit 1
+[ -z "$DRY_RUN" ] && git checkout ${ISCSC_REMOTE}/main
```
Turtyo, line 127: Why remove this `|| exit 1` ? In the case where we are not in a dry run, it will prevent from running the entire script

Turtyo (approved):

scripts/bump.sh
ctmbl: But that's not what we want!
this was a mistake I added in the previous patch, it should be surrounded by brackets because it should applied on the exit code of `git checkout ..`, but I just realized that another test at the beginning already took care of this case so this shouldn't happen!
if you wanna be sure just test the branch without this commit (which should be easy once you cloned it just `git rebase --onto main -i` (something like this))

PR #71, opened 2023-03-11
Add subcommands to bump script

ctmbl:
This PR removes the old way of using `bump.sh`, e.g. `./scripts/bump.sh <version>` and add three new flags to bump either to patch, minor or major version! with `--patch/-p` `--minor/-m` or `--major/-M`. It also wraps everything into functions for a cleaner script and adds a test that checks that the most recent git tag and the current version are compatible.

### DIFF
--- a/scripts/bump.sh
+++ b/scripts/bump.sh
@@ -139,7 +139,7 @@ git_setup () {
 	# ...and checkout on main to create a version bump branch
 	# (working dir is empty check passed)
 	log_info "Checkout on ${iscsc_remote_name}/main"
-	[ -z "$DRY_RUN" ] && { git checkout ${iscsc_remote_name}/main || exit 1; }
+	[ -z "$DRY_RUN" ] && { git checkout ${iscsc_remote_name}/main --detach || exit 1; }
 	log_info "switching to ${bump_branch}"
 	[ -z "$DRY_RUN" ] && { git switch -c ${bump_branch} || exit 1; }
 }
--- a/scripts/bump.sh
+++ b/scripts/bump.sh
@@ -129,8 +129,8 @@ check_iscsc_remote () {
 # --------------------------------- Git setup ----------------------------------
 
 get_remote () {
-	git remote -v | grep "${SSH_REMOTE}" >/dev/null && { echo "${SSH_REMOTE}"; exit; }
-	git remote -v | grep "${HTTPS_REMOTE}" >/dev/null && { echo "${HTTPS_REMOTE}"; exit; }
+	git remote -v | grep --ignore-case "${SSH_REMOTE}" >/dev/null && { echo "${SSH_REMOTE}"; exit; }
+	git remote -v | grep --ignore-case "${HTTPS_REMOTE}" >/dev/null && { echo "${HTTPS_REMOTE}"; exit; }
 }
 
 git_setup () {
@@ -260,7 +260,7 @@ main () {
 	check_iscsc_remote "${ISCSC_REMOTE}"
 
 	# Define iScsc remote name if it exists and log new version
-	ISCSC_REMOTE_NAME=$(git remote -v | grep "${ISCSC_REMOTE}" | awk '{print $1}' | head --lines 1)
+	ISCSC_REMOTE_NAME=$(git remote -v | grep --ignore-case "${ISCSC_REMOTE}" | awk '{print $1}' | head --lines 1)
 	log_info "'${NEW_VERSION}'>'${CURRENT_VERSION}', '${NEW_VERSION}' is accepted as new version."
 
 	# Setup git branch
--- a/scripts/bump.sh
+++ b/scripts/bump.sh
@@ -71,6 +71,11 @@ check_clean_git_working_dir () {
 # Check that supplied version is semantically correct
 check_version_semantics () {
 	local version="$1"
+	if [ -z "${version}" ]; then
+		log_error "You provided a wrong argument, version can't be computed"
+		usage
+		exit 1
+	fi
 	if [ "${version}" != "$(semver ${version})" ]; then
 		log_error "${version} is not a valid version number according to semver"
 		exit 1
--- a/scripts/bump.sh
+++ b/scripts/bump.sh
@@ -22,6 +22,10 @@ log_warning () { log_any "33" "~" "$1"; }
 log_hint    () { log_any "34" "?" "$1"; }
 log_info    () { log_any "35" "i" "$1"; }
 
+break_down_version () {
+    echo ${1//./ }
+}
+
 # -------------------------------- Basic Checks --------------------------------
 
 # Check that current directory is repo root
--- a/scripts/bump.sh
+++ b/scripts/bump.sh
@@ -237,8 +237,8 @@ main () {
 	check_clean_git_working_dir
 
 	# Define version variables
-	NEW_VERSION=$(bump_${COMMAND} ${CURRENT_VERSION})
 	CURRENT_VERSION=$(npm pkg get version | sed 's/"//g')
+	NEW_VERSION=$(bump_${COMMAND} ${CURRENT_VERSION})
 
 	# Define git variables
 	BUMP_BRANCH="${NEW_VERSION}-version-bump"
--- a/scripts/bump.sh
+++ b/scripts/bump.sh
@@ -9,8 +9,6 @@ DEPENDENCIES=(
 # Variables
 NB_ARGS="$#"
 USAGE="bump.sh [-h | --help] [-p | --patch] [-m | --minor] [-M | --major]"
-HTTPS_REMOTE='https://github.com/iScsc/iscsc.fr.git'
-SSH_REMOTE='git@github.com:iScsc/iscsc.fr.git'
 
 # ------------------------------- Tool functions -------------------------------
 
@@ -123,26 +121,19 @@ check_iscsc_remote () {
 	if [ -z "$remote" ]; then
 		log_error "'iScsc/iscsc.fr' remote is not in remote list"
 		exit 1
-	elif [ "$remote" == "${HTTPS_REMOTE}" ]; then
-		log_warning "Using HTTPS remote, SSH remote not found"
 	fi
 	log_ok "'iScsc/iscsc.fr' is in remote list as '$remote' OK"
 }
 
 # --------------------------------- Git setup ----------------------------------
 
-get_remote () {
-	git remote -v | grep --ignore-case "${SSH_REMOTE}" >/dev/null && { echo "${SSH_REMOTE}"; exit; }
-	git remote -v | grep --ignore-case "${HTTPS_REMOTE}" >/dev/null && { echo "${HTTPS_REMOTE}"; exit; }
-}
-
 git_setup () {
-	local iscsc_remote_name="$1"
+	local iscsc_remote="$1"
 	local bump_branch="$2"
 	# ...and checkout on main to create a version bump branch
 	# (working dir is empty check passed)
-	log_info "Checkout on ${iscsc_remote_name}/main"
-	[ -z "$DRY_RUN" ] && { git checkout ${iscsc_remote_name}/main --detach || exit 1; }
+	log_info "Checkout on ${iscsc_remote}/main"
+	[ -z "$DRY_RUN" ] && { git checkout ${iscsc_remote}/main --detach || exit 1; }
 	log_info "switching to ${bump_branch}"
 	[ -z "$DRY_RUN" ] && { git switch -c ${bump_branch} || exit 1; }
 }
@@ -254,7 +245,7 @@ main () {
 
 	# Define git variables
 	BUMP_BRANCH="${NEW_VERSION}-version-bump"
-	ISCSC_REMOTE=$(get_remote)
+	ISCSC_REMOTE=$(git remote -v | grep 'git@github.com:iScsc/iscsc.fr.git' | awk '{print $1}' | head --lines 1)
 
 	# Run all advanced checks
 	check_version_semantics "${NEW_VERSION}"
@@ -262,12 +253,10 @@ main () {
 	check_version_greater
 	check_iscsc_remote "${ISCSC_REMOTE}"
 
-	# Define iScsc remote name if it exists and log new version
-	ISCSC_REMOTE_NAME=$(git remote -v | grep --ignore-case "${ISCSC_REMOTE}" | awk '{print $1}' | head --lines 1)
 	log_info "'${NEW_VERSION}'>'${CURRENT_VERSION}', '${NEW_VERSION}' is accepted as new version."
 
 	# Setup git branch
-	git_setup ${ISCSC_REMOTE_NAME} ${BUMP_BRANCH}
+	git_setup ${ISCSC_REMOTE} ${BUMP_BRANCH}
 
 	# Bump
 	bump_modules ${NEW_VERSION}
@@ -275,8 +264,8 @@ main () {
 
 	# Try to push bump refs
 	log_info 'Pushing branch and bump commit'
-	log_warning "pushing to \`${ISCSC_REMOTE_NAME}\` please type your passphrase/password if required:"
-	PUSH_COMMAND="git push ${ISCSC_REMOTE_NAME} ${BUMP_BRANCH} v${CURRENT_VERSION}"
+	log_warning "pushing to \`${ISCSC_REMOTE}\` please type your passphrase/password if required:"
+	PUSH_COMMAND="git push ${ISCSC_REMOTE} ${BUMP_BRANCH} v${CURRENT_VERSION}"
 	[ -z "$DRY_RUN" ] && { $PUSH_COMMAND || log_error "push failed, you can push with \`${PUSH_COMMAND}\`"; }
 
 	log_hint '`npm install` has been run during the bump, you MUST review the changes during PR review to ensure package.json and package-lock.json where compatible!!!'
--- a/scripts/bump.sh
+++ b/scripts/bump.sh
@@ -92,6 +92,8 @@ check_version_and_tag_consistency () {
 	local json_version="$1"
 	tagged_commit=$(git rev-list --tags --max-count=1)
 	tag_version=$(git describe --tags $tagged_commit)
+	# remove the 'v' before the version
+	tag_version=$(semver $tag_version)
 
 	if [ "$json_version" != "$tag_version" ]; then
 		log_error "tag '$tag_version' is not consistent with project version '${json_version}' !!!"
--- a/scripts/bump.sh
+++ b/scripts/bump.sh
@@ -116,10 +116,12 @@ check_version_greater () {
 check_iscsc_remote () {
 	local remote="$1"
 	if [ -z "$remote" ]; then
-		log_error "'iScsc/iscsc.fr' remote is not in remote list"
-		exit 1
+		log_warning "'git@github.com:iScsc/iscsc.fr.git' remote is not in remote list, won't push"
+		PUSH="n"
+	else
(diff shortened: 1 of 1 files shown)


scripts/bump.sh
```diff
@@ -21,9 +22,13 @@ log_warning () { log_any "33" "~" "$1"; }
 log_hint    () { log_any "34" "?" "$1"; }
 log_info    () { log_any "35" "i" "$1"; }
 
+break_down_version () {
```
amtoine, line 25: no idea what this does in `bash` :thinking:

scripts/bump.sh
```diff
@@ -113,45 +122,154 @@ check_iscsc_remote () {
...
+	check_arg "${NB_ARGS}"
+	check_dependencies
+	check_clean_git_working_dir
+
+	# Define version variables
+	NEW_VERSION=""
+	CURRENT_VERSION=$(npm pkg get version | sed 's/"//g')
+
+	# Extract new version from arguments
+	while [[ $# -gt 0 ]]; do
```
amtoine, line 231: here i would fully parse the arguments before doing anything!

for instance, running `./script/bump.sh --help` does run `npm` and stuff happens...
this is not what i expect for a `--help` :thinking: 

what i would do is
- put the `while` at the top of the `main`
- instead of running `NEW_VERSION=$(bump...)` you can store the action to be performed
- you execute `check_pwd`, etc, ...
- finally you switch over the action value and run the appropriate `bump_*`

amtoine (changes requested):
that looks a lot better, thanks!

i'm gonna try the script a bit, at least, apart from the minor remarks i had, the source looks great :+1:

amtoine:
mm @ctmbl 

i run either one of the following from clean work tree
```
./scripts/bump.sh --patch
```
```
./scripts/bump.sh --minor
```
```
./scripts/bump.sh --major
```
and i get 
```
[i] Current version is '0.1.1'
[-] 'iScsc/iscsc.fr' remote is not in remote list
```
but it really looks to be in my remote list :confused: 
```bash
>_ git remote -v
iscsc	https://github.com/iScsc/iscsc.fr (fetch)
iscsc	https://github.com/iScsc/iscsc.fr (push)
origin	ssh://git@github.com/amtoine/iscsc.fr (fetch)
origin	ssh://git@github.com/amtoine/iscsc.fr (push)
```

scripts/bump.sh
```diff
@@ -8,6 +8,7 @@ DEPENDENCIES=(
 
 # Variables
 NB_ARGS="$#"
+USAGE="bump.sh [-h] -p|-m|-M"
```
amtoine, line 11: as with `git`, i would find
```
bump.sh [-h | --help] [-p | --patch] [-m | --minor] [-M | --major]
```
quite nice and clear

scripts/bump.sh
ctmbl: What do you mean? it actually works and you're the one that give me that piece of code a few months ago :laughing:

scripts/bump.sh
ctmbl: Actually it looks nice I agree but I wanted to make it clear that one of these arguments had to be given!

ctmbl:
> and i get
[i] Current version is '0.1.1'
[-] 'iScsc/iscsc.fr' remote is not in remote list

>but it really looks to be in my remote list 
>$ git remote -v
iscsc	https://github.com/iScsc/iscsc.fr (fetch)
iscsc	https://github.com/iScsc/iscsc.fr (push)
origin	ssh://git@github.com/amtoine/iscsc.fr (fetch)
origin	ssh://git@github.com/amtoine/iscsc.fr (push)

>EDIT: should iScsc/iscsc.fr be called origin?
if so there error message is misleading wink

Thanks for raising this issue to me! Actually if you look closely at the code it doesn't expect iscsc remote from being called `origin` which would be quite unrealiable! But it clearly expect the remote of being a ssh remote and not a https one :smile: 
Do you think I should change that and accept both? or maybe just raise a warning and don't try to push but let the script run?

scripts/bump.sh
ctmbl: that's a really good remark I'll change that! tks!

scripts/bump.sh
amtoine: > Actually it looks nice I agree but I wanted to make it clear that one of these arguments had to be given!
>
yup, that's an issue i have with the current usage i think...

> Have you something in mind to express that? If not no worries I'll just replace USAGE with your proposal!
>
we should find a usage that express "one and only one option has to be given" :thinking: 

i do not really know tbh :eyes:

amtoine:
> Thanks for raising this issue to me! Actually if you look closely at the code it doesn't expect iscsc remote from being called `origin` which would be quite unrealiable!
>
ok perfect :relieved: 

> But it clearly expect the remote of being a ssh remote and not a https one smile Do you think I should change that and accept both? or maybe just raise a warning and don't try to push but let the script run?
>
mm i'd say at least say what the exact problem is :+1: 

whether to be strict about the `git` protocol, i'm not sure :thinking: 
the warning message is a priority, at least to let the user change the remote by hand :wink:

scripts/bump.sh
ctmbl: done!

scripts/bump.sh
ctmbl: > we should find a usage that express "one and only one option has to be given" thinking

Actually `bump.sh -p|-m|-M` expresses that, I think?

ctmbl:
@amtoine 
I improved the log message and avoid the push step in case the remote is not found, tell me what you think of that!
I also went with your USAGE proposal.

scripts/bump.sh
amtoine: :eyes: 

> it actually works
>
never a good reason not to give an example for some obscur `bash` oneline right? :smirk: 

> and you're the one that give me that piece of code a few months ago
>
:grimacing: 
do not remember that and i do not use `bash` anymore, so...

scripts/bump.sh
amtoine: but you changed to my proposition right? :thinking:

amtoine: ok saw that in your comment below :+1:

atxr (approved):
Read the code and tested the new script.
It is indeed easier to use that before!
I don't really have any comments on the code :racehorse:

However I have few concerns when reading the output I get:
```
^(9fae4dd) λ ./scripts/bump.sh -m                                                                                                                                     (iscsc.fr) 20:21:24
[i] Current version is '0.1.1'
[~] 'git@github.com:iScsc/iscsc.fr.git' remote is not in remote list, won't push
[i] '0.2.0'>'0.1.1', '0.2.0' is accepted as new version.
[i] Checkout on 0.2.0-version-bump/main
error: pathspec '0.2.0-version-bump/main' did not match any file(s) known to git
[i] switching to
error: switch `c' requires a value
[i] Bumping `frontend`

> version
> npm install && git add package.json package-lock.json


up to date, audited 1490 packages in 3s

296 packages are looking for funding
  run `npm fund` for details

6 high severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
v0.2.0
[i] Bumping `backend`

> version
> npm install && git add package.json package-lock.json


up to date, audited 154 packages in 762ms

14 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
v0.2.0
[i] Bumping `root`

> version
> git add package.json

v0.2.0
[detached HEAD 348cea8] Bump to version 0.2.0
 5 files changed, 7 insertions(+), 7 deletions(-)
[?] `npm install` has been run during the bump, you MUST review the changes during PR review to ensure package.json and package-lock.json where compatible!!!
```

amtoine:
@atxr 
do you approve or do you have a question? :yum:

scripts/bump.sh
ctmbl: > never a good reason not to give an example for some obscur bash oneline right? 

Right but actually that's not what you asked in the first time. It's simply a find/replace like you would do with `sed` or in `vim`:
```
V=1.2.3
echo ${V//./ } # 1 2 3
```

>  do not remember that and i do not use bash anymore, so...

Doesn't mean you forgot everything, does it?

ctmbl:
@atxr 
> Are these errors meaningful for you?

Oh actually yes I get it!
As @amtoine highlighted it before the problem comes from the script not finding the `iScsc` remote, you can see the warning above:
```[~] 'git@github.com:iScsc/iscsc.fr.git' remote is not in remote list, won't push```
Before last commit this would end up exiting the script but I changed it thinking that stepping over the push stage would be sufficient, but I was wrong this is not the only time the variable `ISCSC_REMOTE` is used. It is in the `git_setup` function and because it is empty, the 2nd argument `BUMP_BRANCH` is interpreted as the first one and here we understand these confusing error message.
I'll just revert last commit to exit the script in case the `ISCSC_REMOTE` is empty, and allow `https` remotes @amtoine

**Comment by ctmbl** [comment]
@amtoine
I also noted something strange:
> $_ git remote -v
iscsc	https://github.com/iScsc/iscsc.fr (fetch)
iscsc	https://github.com/iScsc/iscsc.fr (push)
origin	ssh://git@github.com/amtoine/iscsc.fr (fetch)
origin	ssh://git@github.com/amtoine/iscsc.fr (push)

Did you modified your `git remote -v` output?
I get 
```
$ git remote -v
fork	git@github.com:ctmbl/iscsc.fr.git (fetch)
fork	git@github.com:ctmbl/iscsc.fr.git (push)
origin	git@github.com:iScsc/iscsc.fr.git (fetch)
origin	git@github.com:iScsc/iscsc.fr.git (push)
```
no `ssh://` before my ssh cloned remote and `.git` at the end
Note that this is what GitHub provides when cloning https://github.com/iScsc/iscsc.fr (`Code` green button)

**Comment by atxr** [comment]
It's weird because I'm still getting an error even if iScsc/iscsc.fr is in the remote list :thinking: 
```
[-] 'iScsc/iscsc.fr' remote is not in remote list
```

File: scripts/bump.sh [hunk 4]
```diff
@@ -8,6 +8,9 @@ DEPENDENCIES=(
 
 # Variables
 NB_ARGS="$#"
+USAGE="bump.sh [-h | --help] [-p | --patch] [-m | --minor] [-M | --major]"
+HTTPS_REMOTE='https://github.com/iScsc/iscsc.fr.git'
```
atxr, line 12: I found out why.
In my git remote, urls are with a lowercase `s`
I don't know if it's only for me, but we should do something about this
```suggestion
HTTPS_REMOTE='https://github.com/iscsc/iscsc.fr.git'
```
Another way could be to lowercase the url before the check

scripts/bump.sh
```diff
@@ -8,6 +8,9 @@ DEPENDENCIES=(
...
+SSH_REMOTE='git@github.com:iScsc/iscsc.fr.git'
```
atxr, line 13:
```suggestion
SSH_REMOTE='git@github.com:iscsc/iscsc.fr.git'
```

atxr (changes requested):
I spotted why :bug:

scripts/bump.sh
ctmbl: or just use `grep --ignore-case` maybe? however that's weird... we should have the same git and remotes?

ctmbl:
> It's weird because I'm still getting an error even if iScsc/iscsc.fr is in the remote list thinking

At least it's better than messing up with git on the wrong branch!

scripts/bump.sh
atxr: Yes perfect!

ctmbl: fixed :wink:

scripts/bump.sh
```diff
@@ -67,32 +72,38 @@ check_clean_git_working_dir () {
...
-BUMP_BRANCH="${NEW_VERSION}-version-bump"
-ISCSC_REMOTE=$(git remote -v | grep 'git@github.com:iScsc/iscsc.fr.git' | awk '{print $1}' | head --lines 1)
-
 # ------------------------------ Advanced checks -------------------------------
 
 # Check that supplied version is semantically correct
 check_version_semantics () {
 	local version="$1"
+	if [ -z "${version}" ]; then
+		log_error "You provided a wrong argument, version can't be computed"
```
Turtyo, line 81: Maybe "You didn't provide the version" would be a bit more clear than "wrong argument", because a wrong argument can be a number of things

scripts/bump.sh
```diff
@@ -109,49 +120,168 @@ check_iscsc_remote () {
 	if [ -z "$remote" ]; then
 		log_error "'iScsc/iscsc.fr' remote is not in remote list"
 		exit 1
+	elif [ "$remote" == "${HTTPS_REMOTE}" ]; then
+		log_warning "Using HTTPS remote, SSH remote not found"
 	fi
 	log_ok "'iScsc/iscsc.fr' is in remote list as '$remote' OK"
```
Turtyo, line 126: Since I see that you changed iScsc to a lower case (as per Atxr's request), maybe the log should be consistent with this change

scripts/bump.sh
```diff
@@ -109,49 +120,168 @@ check_iscsc_remote () {
...
+	done
+
+	# Run all basic checks
+	check_pwd
+	check_arg "${NB_ARGS}"
+	check_dependencies
+	check_clean_git_working_dir
+
+	# Define version variables
+	CURRENT_VERSION=$(npm pkg get version | sed 's/"//g')
```
Turtyo, line 249: maybe a little comment on this `sed` expression could be welcomed

Turtyo (commented):
Not much to say about the code, apart from the fact that it looks a lot clearer now. It's mostly comments about code clarity in some parts.

scripts/bump.sh
ctmbl: Actually with this PR the argument is no longer a version, but a flag! Then if you don't provide either of `-p|-m|-M` `NEW_VERSION` is empty and we have this error 😉

scripts/bump.sh
ctmbl: I didn't 😉 I just added `--ignore-case` toi greps, anyway in this log I refer to the protocol, thus the upper case!

scripts/bump.sh
ctmbl: I could but it's a pretty basic one! juste remove every `"` encountered

scripts/bump.sh
Turtyo: Ah yes I forgot that the version was parsed from the argument, it makes sense.

scripts/bump.sh
Turtyo: What is the goal of this operation then ?

ctmbl: If you try yourself `npm pkg get version` you'll see that the ouput looks like `"0.1.1"` this is really annoying I don't why... but anyway I have to get rid of the quotes for the rest of the script

Turtyo: Ok I see

Turtyo (approved):

amtoine:
@ctmbl 

i added the remote as they show up in the `git remote -v`
i think

i don't know, it's quite old now :grimacing:

in the end, i think the `ssh://` is implicit when you have `git@github.com`, i'd say `git` knows about that :+1:

scripts/bump.sh
ctmbl: > as it's not a sed command nor a vim command, i'd like to have the example you gave in the docstring of break_down_version 😋

I can do that!
However maybe not tonight, could you approve if there isn't anything else on your mind? to avoid delaying this PR 24h more ?

amtoine: well i dunno, maybe i'll have something to say after the change :shrug:

ctmbl:
```suggestion
break_down_version () {
# V=1.2.3
# echo ${V//./ } # 1 2 3
``` 
something like that would be sufficient?

amtoine: i'd just add
```
# # Example
# <your example>
```
but otherwise that's good :ok_hand:

ctmbl:
@amtoine 
> in the end, i think the ssh:// is implicit when you have git@github.com, i'd say git knows about that 👍

Actually that wasn't my point at all, and I'm much more worried about the `.git` that you haven't
If we have different `git remote -v` the `grep` won't work that's the problem...

scripts/bump.sh
ctmbl:
```suggestion
break_down_version () {
# # Exemple
# V=1.2.3
# echo ${V//./ } # 1 2 3
```

ctmbl: actually when I clicked "Commit suggestion" it automatically resolved the thread tell me if it's ok @amtoine

amtoine: not sure about the indentation but eh that's fine :+1:

ctmbl: euh yeah you're right, i'll fix it sometime 🙂

amtoine:
@ctmbl 
> If we have different git remote -v the grep won't work that's the problem...
>
you can try to match for any remote containing `iscsc/iscsc.fr`.
or you can just `git remote add bump <the remote you want>` and with that instead and never use `git remote -v`.
that might be the easiest :+1:

> So why do our outputs are f*cking different?
>
you can `git remote add` anything, both for the remote name and the remote location.
that won't be a problem for `git` until you actually try to interact with upstream.

e.g.
```bash
git remote add djwaldj djhcklnzsc
```
would work just fine

ctmbl:
@amtoine 
## Answers
> you can try to match for any remote containing iscsc/iscsc.fr.

I was thinking of that, but that doesn't solve the upper/lower case problem, it has been managed by `--ignore-case` flag added to each `grep` but I don't like it... quite error-prone while maintaining the script.
However it does solve other problems
##
 >  or you can just git remote add bump <the remote you want> and with that instead and never use git remote -v.
 that might be the easiest 👍 
 
 The easiest yes but here are the reasons why I need this remote:
  - checkout on `<ìscsc remote>/main` (assuming that if the user is actually bumping it is an active maintainer --> this ref has good chances to be up-to-date)
  - push `<version>-version-bump` to `<iscsc remote>`
 
While the 2nd step isn't necessary I find the first one quite mandatory, then:
If I was to follow your solution (adding a `bump` remote) I'd have to:
 - add the remote (messing up with the user remotes...)
 - fetch (adding a new maybe unwelcomed and redundant ref to the user tree)
 - checkout, switch (and push *(optional step)*) (but here it will work perfectly)
 - remove the fetched ref ???
 - remove the new remote ???

Quite an overhead in my opinion, I'd prefer to use the existing remotes and let the user deal with them in the way he likes it 👍

##

>you can git remote add anything, both for the remote name and the remote location.
that won't be a problem for git until you actually try to interact with upstream.

Yes ofc but I don't see how this is relevant... or **did you** add garbage remotes? I implicitely assumed that what was in **your** remotes worked, then, why does it work with different remotes? didn't you just copy/paste the remotes given by GitHub when cloning or adding new remotes?
Well in any case it works so I have to deal with it 👍 

## Summary

 For the reasons stated in my 2nd answer I think I'll go with the existing remotes in the user remotes, I'll just have to carefully deal with edge cases as you have.
 The good thing is, we're three and we have three different remotes, it probably gives me every different possibilities the script may encounter and this is useful to highlight bugs now and not when this PR will be merged, so even if it pisses me off thanks guys 👍

amtoine (approved):

ctmbl:
> these are not "garbage" remotes, they work just fine

Ofc they aren't in your case because you're using them! I was caricaturing because you previously showed a garbage one

>    ^ssh:// is optional when you already have git@github.com
    .git$ is optional

Ok this is interesting, annoyting and cool at the same time I should say

atxr:
Thanks for solving this bug @ctmbl 
You can safely merge now :wink:

Issue #64, opened 2022-12-16
Add containerization in development mode


ctmbl:
#20 added containerization to deploy the website in production mode. 

I would like to extend it to development mode in order to:
 - not mess with the host (developer)'s environment
 - solve env issues for Windows developer
 - homogenize modes
 - create a single script to run either development or production mode, transparently for the user 
 - solve potential version/environment problem for developer working on different machines
 
 Would be nice to have feedback on this!

 *I'm currently working on so I'm assigning it to myself*

amtoine:
> * not mess with the host (developer)'s environment
> * solve env issues for Windows developer
> * homogenize modes
> * create a single script to run either development or production mode, transparently for the user
> * solve potential version/environment problem for developer working on different machines
> 
looks great :ok_hand: 

> _I'm currently working on so I'm assigning it to myself_
>
a preview of what you would like to do? :yum: 
i feel this would require enhancing the `Dockerfile`s, right? :thinking:

ctmbl:
@amtoine 
> a preview of what you would like to do? yum

I see it that way: currently to start in dev mode you must open 2 terminals and run the needed commands in the `./backend` folder first, then `./frontend`. 
The two process communicate together via the opened ports (defined in the `.env.development`) and the frontend calls the backend endpoints (`/api/*`) prepending the URL defined in the `./frontend/package.json` : 
```json
{"proxy": "http://localhost:3001"}
```
 @atxr will correct me if I'm wrong

Now I want to put these two independent services in two containers, as done in production mode but to communicate together `localhost` won't do. 
I have some solutions:
 - change the proxy to `http://<backend container>:<port>` (like [here](https://stackoverflow.com/a/50207462)) which I'd like to avoid --> starting dev mode without docker won't work anymore
 - define the proxy in the code, making it depending of an an env var --> didn't try it but I [saw this post](https://stackoverflow.com/a/70413065)
 - use an `nginx` container without HTTPS as done in prod mode --> will require another nginx conf
 
Don't know what's the best solution so I'm listening to your opinion. 
Currently it's the only hard change needed I identified, another one is the `start` script in the same file but it's easily solvable.
 
>  i feel this would require enhancing the Dockerfiles, right?

No a different one instead!

amtoine:
@ctmbl
thanks for the reply :yum: 

it's a bit technical for me right now :confused: 
i did not follow everything about `docker` here :eyes: 

[...]

PR #83, opened 2023-03-30
add a CI to format the code

### LINKED ISSUE #82: add a CI to check the format of the code
cc/ @atxr 

about the question in #78 and the *change request* in #81.

@atxr, could you add a very simple CI in `.github/workflows/` to ~format~ check the format of the code base with `prettier`? :yum: 

**EDIT**: the CI should fail if the format is not good. the CI should not try to commit on top of a PR branch if the format is bad.

atxr:
Add github workflow to format code when pushing on main

Fixes #82


.github/workflow/prettier.yml
```diff
@@ -0,0 +1,25 @@
+name: Continuous Integration
```
ctmbl, line 1: You could find a better name here I think!

.github/workflow/prettier.yml
```diff
@@ -0,0 +1,25 @@
...
+
+# This action works with pull requests and pushes
+on:
+  pull_request:
+  push:
+    branches:
+      - main
```
ctmbl, line 8: To see it in action could you temporarily set up to 
```suggestion
  pull_request:
    branches:
      - main
  push:
    branches:
      - main
```
See (after the commit) should force the action to be ran on this PR (we will see the output in the Actions tab
And we'll be able (maybe?) to understand its output (in order to see how it handles the commiting part of the formatting :+1: )

.github/workflow/prettier.yml
```diff
@@ -0,0 +1,25 @@
...
+        uses: actions/checkout@v3
+        with:
+          # Make sure the actual branch is checked out when running on pull requests
+          ref: ${{ github.head_ref }}
+
+      - name: Prettify code
+        uses: creyD/prettier_action@v4.2
+        with:
+          # This part is also where you can pass other options, for example:
+          prettier_options: --write **/*.{js,md}
```
ctmbl, line 25: as said in private maybe add 
`commit_options`
or `push_commit`
or `same_commit`
see https://github.com/creyD/prettier_action

ctmbl (changes requested):
Quite some questions before approving this but we're on the right track!

.github/workflow/prettier.yml
atxr: I can do it if you want, but I tested it on another repo and it works perfectly
The output (if a none formatted file is found) is something like
```
[file.js] Error: Did you forget to run prettier on this file?
```

.github/workflow/prettier.yml
atxr: I added `check` instead of write, to avoid modifying the code automatically, which can be a bit weird.
The action will block the PR if the code isn't formatted correctly

atxr:
Sorry for the force push, it was to resolve conflicts

.github/workflow/prettier.yml
ctmbl: Ok that's wonderful but does it tell the user what the error is?
Maybe somebody won't want to install prettier globally (and currently we don't have a dev env, and #64 won't introduce such possibilities I think)

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3.8'
+version: "3.8"
```
ctmbl, line 1: this is not a `**/*.js|md` file why had it been formatted?

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          # Make sure the actual branch is checked out when running on pull requests
+          ref: ${{ github.head_ref }}
+
+      - name: Prettify code
+        uses: creyD/prettier_action@v4.3
+        with:
+          # This part is also where you can pass other options, for example:
+          dry: True
+          prettier_options: --check "**/*.{js,md,yml,css,html}"
```
ctmbl, line 26: are these options redundant?
And I see that you added `yml` and other files extensiosn, which maybe answer my other question

ctmbl (commented):
Quite impossible to review everything I must confess, but still I have some little questions for you!

.github/workflow/prettier.yml
atxr: You mean the error should give a way to fix the formatting?
Mmm if one of the contributors doesn't want to install npm then it's a little bit tricky to agree with the formatting.
You can install it locally if you don't want it globally 🤔 
And I can definitely make some doc if needed.

atxr:
I understand I modified many things but if you want to take a close look you can:
- check the prettier config files (I think there are the same for backend and frontend)
- try to understand the github action and see if we agree on it (regarding the opened discussion, there is still some work here)

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,26 @@
...
+      - main
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
```
amtoine, line 17: i think this `with` block is not mandatory.
`actions/checkout@v3` should arealy checkout on top of the PR branch :thinking: 

the best example i have for this is the [`std-lib-and-python-virtualenv` job of the `nushell` CI](https://github.com/nushell/nushell/blob/main/.github/workflows/ci.yml#L85), 'cause i partially wrote it myself :eyes:

.github/workflows/prettier.yml
amtoine: i think @ctmbl meant `dry: True` and `--check` might be redundant?

amtoine (commented):
we cannot have a top-level `prettier.config.yml` file, instead of two file in `backend/` and `frontend/`?
or maybe you plan to apply multiple different rules?

docker-compose.yml
amtoine: agree with @atxr

.github/workflow/prettier.yml
amtoine: so please do not change the `on:` field @atxr.

it should run as-is :+1:

atxr:
I think it doesn't run on PR's for now, we need to push it first

.github/workflows/prettier.yml
amtoine: mm okok :+1:

amtoine:
> I think it doesn't run on PR's for now, we need to push it first
>
that is strange.
how do you test a CI before merging?
it could be very broken and block all the new PRs on top of `main` :thinking: 

and i remember CIs running as soon as they have a `.github/workflows/*.yml` file on top of the PR branch :thinking:

atxr:
Ok! Here is the **UPDATE**
The CI doesn't run because I'm working on a branch in [my fork](https://github.com/atxr/iscsc.fr)
I tested firstly on a dummy repo that I deleted, and now, I opened a dummy PR on my fork to test the CI, and it works! :partying_face: 

So as soon as it will be merged here, it will start don't worry @amtoine !
I also solved all the discussion above with the last test, but I'll let @ctmbl close them

Finally, I did a test with a none formatted file to see if the CI fails, and it worked.
I add some trouble then when reverting the commit (because I was working on the sh*tty vscode online version that auto-format on save with a different prettier config) but now everything is ready to be merged.

.github/workflows/prettier.yml
atxr: Yes indeed, I removed it :heavy_check_mark:

ctmbl: That was it tks!

ctmbl: Indeed I prefer to use raw prettier argument to pass to it, instead of the [creyD prettier_action](https://github.com/creyD/prettier_action). Tks for [6087ce7](https://github.com/iScsc/iscsc.fr/pull/83/commits/6087ce7680edc414a6be1634101d1a2dd7e33306)

docker-compose.yml
ctmbl: Seems good!

ctmbl:
I think someone should mention that this action comes from https://github.com/creyD/prettier_action

.github/workflow/prettier.yml
ctmbl: > so please do not change the on: field @atxr.

@amtoine 
You actually got me wrong here... It wasn't a request to change the `on:` field *(even if I believe that running this `on: pull_request: main` would be a good idea and would have avoided the confusing discussion I see in the comments below...)*

My point here is that the error msg would be far better if it also showed where the formatting error are. As pylint and black do for example!

----------------------

>  Mmm if one of the contributors doesn't want to install npm then it's a little bit tricky to agree with the formatting.

@atxr I quite don't agree with you here... Actually the dev env I'm building in #64 and #86 is designed to avoid installing iscsc.fr dev stuff on your local machine! I see `npm` and `prettier` as such dev tools so maybe I'll write a little script that runs the docker containers and run prettier in it, it might seems over-complicated to you but I like not to mess with my machine, and I'm creating the dev env for this.

>  You can install it locally if you don't want it globally thinking

And this isn't a suited solution as we decided together a few months ago not to install packages locally in root, installing prettier locally would modify the package.json and create a package-lock.json at root :+1:

ctmbl:
> but why doesn't it run? :sweat_smile:

As @atxr discovered and epxlained it after your question @amtoine, it will run once we click the `Squash and merge` button, because the action states:
```yml
on:
  pull_request:
  push:
    branches:
      - main
```
it doesn't trigger on pull request, only on push, a merging **is** a push (ordered by GitHub) so it'll trigger when we'll decide to merge.
This is why I wrote [this](https://github.com/iScsc/iscsc.fr/pull/83/commits/63cdb6418861e316f6fcdd8350e20ef5ca48b9c1#r1154435355) in a thread above.

### Why I find this awful
Imagine a workflow:
 - I cerate a PR
 - you review it
 - I commit modifications
 - You approve 
 - I merge it (finally)
This is a quite long process already, even if I truly believe that it is a good and necessary one.

However after this PR the journey won't be over, we'll add a (not necessary in my opinion) overhead:
because we trigger the action only on push teh CI will fails when trying to merge, rejecting the merge, requiring a new commit, dismissing the approvals of the reviewers (becuase the `Dismiss stale pull request approvals when new commits are pushed` option is ON on this repo's main protection rule [see](https://github.com/iScsc/iscsc.fr/settings/branch_protection_rules/28880568), *and don't get me wrong this is a good policy I want to keep it)

**SO**, I think we should instead trigger the CI on `pull_reequest` **and** `push`
```yml
on:
  pull_request:
    branches:
      - main
  push:
    branches:
      - main
```

looking forward for your POV!

amtoine:
@ctmbl 
i thought
```yaml
pull_request:
```
meant "any pull request" :open_mouth: 

not sure it's very explicit in [the doc here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore) :thinking: 


and again, see the [`on` field of the `ci.yml` of `nushell`](https://github.com/nushell/nushell/blob/main/.github/workflows/ci.yml#L1-L5), it does not define anything in `on.pull_request` yet the CI runs on every single PR :confused: 

i'm actually genuinly confused by this...

amtoine (dismissed):
looking at your fork @atxr, that looks good to me :muscle:

ctmbl:
@amtoine 

>meant "any pull request" 😮

And actually you're right!! (and I was wrong) (and we agree on the trigger events we want, that's truly beautiful)
The only reason that the workflow didn't trigger **here** is thta the file, at the moment (**BEFORE the merging**) only exists on `atxr/iscsc.fr` then can't trigger a workflow here 🥲 
@atxr stated it here:
> The CI doesn't run because I'm working on a branch in [my fork](https://github.com/atxr/iscsc.fr)

but I didn't get it exactly when reading it, now I do!

ctmbl (approved):
the fog and confusion of misunderstanding is gone we're ready to merge 🚀

.github/workflow/prettier.yml
atxr: > My point here is that the error msg would be far better if it also showed where the formatting error are. As pylint and black do for example!

I don't know if such a feature exists, I'll check for you but I didn't find anything like this

> I see `npm` and `prettier` as such dev tools so maybe I'll write a little script that runs the docker containers and run prettier in it, it might seems over-complicated to you but I like not to mess with my machine, and I'm creating the dev env for this.

If you think it's a great move to add it the the dev env, it can be great. However, I don't think we can change anything about this in the scope of this PR

> After looking at #84 I remove this statement, but still some people (like me)don't want to install `npm` at all if possible :+1: 
Again, I totally understand but I think it's related with the dev workflow.

amtoine: it's true that, for instance, `cargo fmt --check` shows where the errors are, which is very nice :+1: 

and for the `npm` installation, i agree with you @atxr => this website is `npm` / `node`-based, i.e. you'll need to have a minimal dev setup.
for a `rust` project, you'd have to install `rustc` and `cargo`.
for a `python` project, you'd have to install `python` and probably `pylint` or `mypy`.
etc, etc...

amtoine: for me, the preview of the changes to be applied is nice, but not that critical.
and that can be addressed in another PR :+1:

atxr: Again, I understand that one doesn't want to install those tools on a local setup, because of conflicts for exemple 🙄
But I think we should merge this PR and add improvement to the dev workflow to fix this

amtoine: agree, let's land this and build upon it :+1: 

@ctmbl are you ok with that?
you can resolve this and merge the PR :ok_hand:

amtoine: ooh no, @atxr please solve the conflict :wink:

atxr:
@ctmbl @amtoine conflict resolved! :heavy_check_mark:

amtoine (approved):
wow that force push was confusing :eyes: 

but looks like before :thinking:

atxr:
Yes sorry I didn't know how to do without a force push that was weird 🙄

.github/workflow/prettier.yml
ctmbl: Yep I agree with both of you these two points:
 - show format error diff in the prettier action logs
 - wrap prettier in the dev env

should be addressed in other PRs
But I wanted to reach a common opinion about these before resolving this thread!

ctmbl (approved):
Looks nice!

amtoine:
> Yes sorry I didn't know how to do without a force push that was weird roll_eyes
>
i think the easiest for authors and the less confusing for reviewers is to
- merge the current conflicting `main` branch, even better the exact revision
- add in the commit message that it's a merge to solve conflicts

that would
- avoid force-pushing
- leave a "linear" history for the reviewers to checkout the new branch without too much intervention
- avoid making all the commits of the PR potentially bad
  - with a merge, you only have to review the new commit
  - with a rebase / force-push, i do not have any guarantee that you did not break anything during the rebase :thinking:

amtoine:
but thanks a lot for the contribution, that's really great :relieved: 

looking forward to seeing the CI catch formatting mistakes :smirk:

ctmbl:
@atxr

I completely agree with @amtoine about merging to solve conflicts due to other PR merges that happened during the reviewing process!

atxr:
Alright, I was always using rebase indeed, but a merge is more appropriate for these cases.

PR #34, opened 2022-11-13
Add https

atxr:
This PR adds HTTPS certificates generation to the deployment pipeline for production mode.
I used `certbot` to generate certificates inside a docker container, with shared volumes with the `nginx` container.
These shared volumes allow `nginx` to load certificates generated by `certbot` inside the configuration and setup ssl.
SSL cipher suits and protocols used are generated by certbot/letsencrypt and are supposed to be the latest one.
This PR will close #33

PR #86, opened 2023-03-30
Create a containerize environment for development workflow

### LINKED ISSUE #13: Automation and enhancement of the dev workflow

### LINKED ISSUE #64: Add containerization in development mode

ctmbl:
Close #64 

This PR adds `Dockerfile`s and `docker-compose` file as well as a nginx conf template for development mode.
The main differences with the prod mode are:
 - shared volume for the source code, it is not embedded in the image
 - no certbot/no HTTPS because we don't need it => a different nginx config
 - not compiled frontend (then less efficient but should refresh itself on code changes)
 - that's it I think?
 
### What solutions I chose
 - `node_modules` are in the image, so they don't overlap with one that could exists on home system, also we don't share the whole frontend and backend folder so local `node_modules` don't overlap with containerized one
 - separated files for dev and prod mode (for Docker, compose, and nginx) --> see next section

### What still needs to be done/doesn't satisfy me
 - [x] the nginx conf simply doesn't work at the time I would greatly appreciate the help of @atxr --> it works see below tks to [00d0912](https://github.com/iScsc/iscsc.fr/pull/86/commits/00d09125ddd061fa805d6f00c4f6462ed56d1428)
 - [x] the backend fails to reach mongoDB I get the error (1) (see Appendices section)
 - [x] the frontend **seems** to fail to reach the backend I get (2)
 - [x] I didn't tested that the website refreshes well on code chnages but I don't doubt that it does because the volumes are shared (cf the `docker-compose-dev.yml`)  --> **tested well on** [fc1c601](https://github.com/iScsc/iscsc.fr/pull/86/commits/fc1c601fcb86a91b356310bdaab8fc888a9a9f08)
 - [x] the call `MODE=$MODE ./run_nginx.sh` in `nginx/Dockerfile` isn't really beautiful I'd say
 - [x] different Dockerfiles or an only one with multiple targets? --> **don't care for now**

### Appendices
(1)
```
iscscfr-node-app-dev-1   | MongooseServerSelectionError: Could not connect to any servers in your MongoDB Atlas cluster. One common reason is that you're trying to access the database from an IP that isn't whitelisted. Make sure your current IP address is on your Atlas cluster's IP whitelist: https://www.mongodb.com/docs/atlas/security-whitelist/
iscscfr-node-app-dev-1   |     at Connection.openUri (/backend/node_modules/mongoose/lib/connection.js:825:32)
iscscfr-node-app-dev-1   |     at /backend/node_modules/mongoose/lib/index.js:414:10
iscscfr-node-app-dev-1   |     at /backend/node_modules/mongoose/lib/helpers/promiseOrCallback.js:41:5
iscscfr-node-app-dev-1   |     at new Promise (<anonymous>)
iscscfr-node-app-dev-1   |     at promiseOrCallback (/backend/node_modules/mongoose/lib/helpers/promiseOrCallback.js:40:10)
iscscfr-node-app-dev-1   |     at Mongoose._promiseOrCallback (/backend/node_modules/mongoose/lib/index.js:1288:10)
iscscfr-node-app-dev-1   |     at Mongoose.connect (/backend/node_modules/mongoose/lib/index.js:413:20)
iscscfr-node-app-dev-1   |     at Object.<anonymous> (/backend/app.js:23:4)
iscscfr-node-app-dev-1   |     at Module._compile (node:internal/modules/cjs/loader:1254:14)
iscscfr-node-app-dev-1   |     at Module._extensions..js (node:internal/modules/cjs/loader:1308:10) {
iscscfr-node-app-dev-1   |   reason: TopologyDescription {
iscscfr-node-app-dev-1   |     type: 'ReplicaSetNoPrimary',
iscscfr-node-app-dev-1   |     servers: Map(3) {
iscscfr-node-app-dev-1   |       'ac-a2ywkbi-shard-00-01.a11re32.mongodb.net:27017' => [ServerDescription],
iscscfr-node-app-dev-1   |       'ac-a2ywkbi-shard-00-02.a11re32.mongodb.net:27017' => [ServerDescription],
iscscfr-node-app-dev-1   |       'ac-a2ywkbi-shard-00-00.a11re32.mongodb.net:27017' => [ServerDescription]
iscscfr-node-app-dev-1   |     },
iscscfr-node-app-dev-1   |     stale: false,
iscscfr-node-app-dev-1   |     compatible: true,
iscscfr-node-app-dev-1   |     heartbeatFrequencyMS: 10000,
iscscfr-node-app-dev-1   |     localThresholdMS: 15,
iscscfr-node-app-dev-1   |     setName: 'atlas-s2e0y3-shard-0',
iscscfr-node-app-dev-1   |     maxElectionId: null,
iscscfr-node-app-dev-1   |     maxSetVersion: null,
iscscfr-node-app-dev-1   |     commonWireVersion: 0,
iscscfr-node-app-dev-1   |     logicalSessionTimeoutMinutes: null
iscscfr-node-app-dev-1   |   },
iscscfr-node-app-dev-1   |   code: undefined
iscscfr-node-app-dev-1   | }
```

(2)
```
iscscfr-react-app-dev-1  | Proxy error: Could not proxy request /api/articles from localhost:3000 to http://localhost:3001.
iscscfr-react-app-dev-1  | See https://nodejs.org/api/errors.html#errors_common_system_errors for more information (ECONNREFUSED).
```


ctmbl:
I was getting with nginx container the following error:
```
iscscfr-nginx-1          | 2023/03/30 13:57:46 [emerg] 9#9: host not found in upstream "react-app" in /etc/nginx/nginx.conf:40
```

.env.example
```diff
@@ -2,9 +2,12 @@
 # development setup #
 #####################
 
+# These are needed for docker-compose
+# Thus we need to specify --env-file .env.development when calling docker-compose
 REACT_PORT=3000  # local port for frontend
 NODE_PORT=3001  # local port for backend
 SERVER_NAME=yourdomainename.com  # your domain name, for the nginx configuration
+MODE=dev|prod # you must chose !
```
amtoine, line 10: what happens if the developer tries to use something not in `dev|prod`?

frontend/package.json
```diff
@@ -20,7 +20,7 @@
     "react-scripts": "5.0.1"
   },
   "scripts": {
-    "start": "env-cmd -f ../.env.development react-scripts start",
+    "start": "react-scripts start",
```
amtoine, line 23: so this is because now this option is on `docker-compose`, right?

nginx/nginx.conf.prod.template
amtoine, line 1: there are some trailing whitespaces on line 49 :thinking: 

they were here before, so it's up to you if you want to remove them or not :wink:

nginx/nginx.conf.dev.template
amtoine, line 1: this looks ok, but why that many dead commented lines?
maybe we can remove them to make the files easier to read?

amtoine (commented):
looks good for now.

solving the issues that make `docker compose` not run properly and i'll approve this :ok_hand:

frontend/package.json
ctmbl: Absolutely! This was needed to load the `.env.development` file (as [dotenv](https://pypi.org/project/python-dotenv/) in python), but is now handled by `docker compose` described in the two `docker-compose-dev.yml`

nginx/nginx.conf.dev.template
ctmbl: Actually we could you're probably right, but as stated in the commit I simply used a previous template of @atxr before #34. This is no good reason to keep these lines but keeping them as example of what could be done seemed fine to me!
However I'm open to different POV

.env.example
ctmbl: It will simply not found the associated template and crash, I must confess I don't know if the error's nice, but in my defense this is configuration stuff, you don't expect a user to writes nonsense here (like for example qtile config crashes and load default in case of semantics error in the config file).

nginx/nginx.conf.prod.template
ctmbl: Hmm I don't see them honestly, could you suggest a change with the '\`\`\`suggestion' github embedding?

.env.example
amtoine: just asking, a crash is far better than unexpected behaviour :relieved:

frontend/package.json
amtoine: thanks :ok_hand:

nginx/nginx.conf.prod.template
amtoine: mm i thought i saw one in the global `git diff` i tried locally, but can not reproduce...

nevermind :wink:

nginx/nginx.conf.dev.template
amtoine: ah ok that's a copy-paste!

i prefer removing them :thinking: 

if you really want to keep them somewhere, maybe add a separate "template / example" file or write a gist? :yum: 

but in the real thing, it makes things confusing i find :thinking:

atxr: At the very beginning, I kept the comment because I was not very friendly with nginx, and it helped me understanding what are the possibility. But know, we can remove them of course.

.env.example
ctmbl: I unresolved this one as a reminder to test this behavior once the dev env works!
This shouldn't be a problem but better be sure :wink:

nginx/nginx.conf.dev.template
ctmbl: Ok np! 
Here I removed unused code [50a6169](https://github.com/iScsc/iscsc.fr/pull/86/commits/50a61690e05562bb181edae63873f10f426dfaf3)
and I also explicitly configured logging  [0c99656](https://github.com/iScsc/iscsc.fr/pull/86/commits/0c99656952ecbe796f4a93fdc1b57555eb77fa1c)

However I'm pretty sure we should add more explicit configuration (instead of relying on default hidden values) for `charset` for example.

amtoine: thanks :yum:

nginx/nginx.conf.dev.template
```diff
@@ -9,6 +11,10 @@ http {
     include mime.types;
     default_type application/octet-stream;
 
+    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                      '$status $body_bytes_sent "$http_referer" '
+                      '"$http_user_agent" "$http_x_forwarded_for"';
+    access_log  logs/access.log  main;
```
atxr, line 17: Same here
```
iscscfr-nginx-1          | 2023/04/08 09:56:33 [emerg] 9#9: open() "/etc/nginx/logs/access.log" failed (2: No such file or directory)
```

nginx/nginx.conf.dev.template
```diff
@@ -1,5 +1,7 @@
 worker_processes 1;
 
+# see https://www.digitalocean.com/community/tutorials/nginx-access-logs-error-logs#nginx-error-log-severity-levels for configuring
+error_log  logs/error.log  info;
```
atxr, line 4: This last commit introduces a bug when trying to lunch the nginx container
```
iscscfr-nginx-1          | nginx: [emerg] open() "/etc/nginx/logs/error.log" failed (2: No such file or directory)
```

dolphinau (commented):
Last commit introduces two errors :thinking:

ctmbl:
This PR will stay idle for a while because we'll be working on two major changes that'll affect it:
 - migrate the DB to a local docker container, that should solve on the bug we currently experience
 - migrate the backend to python

nginx/nginx.conf.dev.template
ctmbl: fixed in [7351f76](https://github.com/iScsc/iscsc.fr/pull/86/commits/7351f76b6a6753f740f427b580e6c39d9c320981)

but as stated in the commit body I'll soon open a PR to use these

nginx/nginx.conf.dev.template
ctmbl: fixed in [7351f76](https://github.com/iScsc/iscsc.fr/pull/86/commits/7351f76b6a6753f740f427b580e6c39d9c320981)

but as stated in the commit body I'll soon open a PR to use these

(same reply)

ctmbl:
This PR is not a draft anymore! @atxr @amtoine

.env.example
ctmbl: didn't test it yet

gbrivady:
Can you add in the `readme` how to launch the container for development mode? My guess is that it isnt much different from production mode, but as of now the `readme` still has the old instructions with  `node` etc...

ctmbl:
@gbrivady good point! tks for noticing it I completely forgot to update the readme, but as you guessed it's really simple juste run `docker-compose --env-file .env.development -f docker-compose-dev.yml up` optionally you can add `-d` for `detach` mode (running in background) and `--build` for rebuilding the images (which will be done anyway if you're running it for the first time)

atxr:
Let's wait for #95 here

gbrivady:
> @gbrivady good point! tks for noticing it I completely forgot to update the readme, but as you guessed it's really simple just run `docker-compose --env-file .env.development -f docker-compose-dev.yml up` optionally you can add `-d` for `detach` mode (running in background) and `--build` for rebuilding the images (which will be done anyway if you're running it for the first time)
> 
> I'll uipdate the README.md really soon!


Got a bunch of errors when I tried your command:
`time="2023-05-04T19:08:13+02:00" level=warning msg="The \"REACT_LOCAL_PORT\" variable is not set. Defaulting to a blank string."
time="2023-05-04T19:08:13+02:00" level=warning msg="The \"REACT_LOCAL_PORT\" variable is not set. Defaulting to a blank string."
time="2023-05-04T19:08:13+02:00" level=warning msg="The \"REACT_LOCAL_PORT\" variable is not set. Defaulting to a blank string."
time="2023-05-04T19:08:13+02:00" level=warning msg="The \"SERVER_NAME\" variable is not set. Defaulting to a blank string."     
time="2023-05-04T19:08:13+02:00" level=warning msg="The \"MODE\" variable is not set. Defaulting to a blank string."
time="2023-05-04T19:08:13+02:00" level=warning msg="The \"DB_PORT\" variable is not set. Defaulting to a blank string."
time="2023-05-04T19:08:13+02:00" level=warning msg="The \"DB_PORT\" variable is not set. Defaulting to a blank string."
time="2023-05-04T19:08:13+02:00" level=warning msg="The \"DB_PORT\" variable is not set. Defaulting to a blank string."
1 error(s) decoding:

* error decoding 'Ports': No port specified: :<empty>
`

I think I might have issues with my .env file; maybe there is a way to fix it without resending me a complete dev one?

atxr:
Yes indeed, looks like the port is missing in your env file
Nevertheless, we should wait for #95 before working on this one because it might change some stuff.

frontend/package.json
```diff
@@ -1,5 +1,4 @@
 {
-  "proxy": "http://localhost:3001",
```
Laudut, line 2: I just want to be sure that it's normal that you delete "proxy"

Laudut (commented):
I don't understand everything but in frontend/package-lock.json half of the time you replace 
 "dev" : true,
with
"devOptionnal" : true,
And the rest of the time you just delete it : it is normal ?

frontend/package.json
ctmbl: yes it is! This proxy definition is too "hardcoded" for us, in particular from the point of view of the frontend container (trying to reach the backend), the backend is not at `https://localhost:3001` it is inside a dedicated docker network *(see the end of the `Dockerfile.dev` and the `network:` section under each container)* and thus needs to be proxied to the right location: `https://react-ap-dev:3001` configurable by environment variables!
This is done by [301f26c](https://github.com/iScsc/iscsc.fr/pull/86/commits/301f26c99bfa1721b0183fadbdd1a1530fd7e38e) at runtime 😉

ctmbl:
> I don't understand everything 

No problem! the point isn't to be an epxert on everything but:
 1- to learn some things
 2- to ask the naive/right questions in order for me to question my choices and know what I did wrong or what's unclear!

> but in frontend/package-lock.json half of the time you replace
"dev" : true,
with
"devOptionnal" : true,
And the rest of the time you just delete it : it is normal ?

Very good question in fact!
I wondered too why this change occurred and lost almost 1hour figuring it out 🥲 
After reading a bit of npm documentation this is simply because for the new dependency I added in [301f26c](https://github.com/iScsc/iscsc.fr/pull/86/commits/301f26c99bfa1721b0183fadbdd1a1530fd7e38e) these subdependencies become Optional, so the mega changes to the `package-lock.json`
Tks for bringing up the issue!

ctmbl:
@Laudut

docker-compose.yml
ctmbl, line 1: not sure if this should go in there... this is a little outside the scope of the PR but still very little changes

ctmbl: the goal is to homogenize both docker compose files so if it's OK for the reviewers I'll let it as is!

.env.example
ctmbl: tks for raising the issue! 
Actually it wasn't working so the website wasn't up on port 80 but teh container wasn't crashing, quite hard to understand the logs. So I armored the script against that: [fc1c601](https://github.com/iScsc/iscsc.fr/pull/86/commits/fc1c601fcb86a91b356310bdaab8fc888a9a9f08) tks again!

atxr:
Will do this evening 🚒

ctmbl:
@atxr tks!

docker-compose.yml
atxr: I don't really understand this conversation, some messages may have been deleted :thinking: 
Let's close it if so :eyes:

frontend/package.json
atxr: @Laudut don't forget to close this conversation!

.env.example
atxr: Here is what happen is you put `dev|prod` with fc1c601
```
[atxr@arch iscsc.fr]$ docker logs iscscfr-nginx-1 
nginx template: 'nginx.conf.dev|prod.template' not found, exiting
```
@amtoine I'll let you close this

dolphinau (dismissed):

gbrivady (dismissed):

dolphinau (dismissed):

gbrivady (dismissed):

scripts/setup-db-folder.sh
```diff
@@ -0,0 +1,45 @@
...
+
+log_error   () { log_any "31" "-" "$1"; }
+log_ok      () { log_any "32" "+" "$1"; }
+log_warning () { log_any "33" "~" "$1"; }
+log_hint    () { log_any "34" "?" "$1"; }
+log_info    () { log_any "35" "i" "$1"; }
+
+check_pwd () {
+  root="$(dirname $(realpath $(dirname "$0")))"
+  [ "$root" != "$(pwd)" ] && {
```
J3y0, line 15: Are you sure you check the right path here ?

I mean, I tested the command locally and here what I found:
```bash
[ iscsc.fr ] on dev-container [?] & dirname $(realpath $(dirname .))   
/home/math/Documents

[ iscsc.fr ] on dev-container [?] & pwd     
/home/math/Documents/iscsc.fr
```
And I am right at the root of the repo. Is it because the script is in another directory `dirname` will take the directory from where the script is and not from where it is launched ? I find it utterly strange there are no other ways of checking this. And yet, I just think about doing a `cd ..` followed by a `pwd`, which is not better you could say...

scripts/setup-db-folder.sh
```diff
@@ -0,0 +1,45 @@
...
+}
+
+setup_folder () {
+    local folder="$1"
+
+    mkdir -p $folder/dev
+    mkdir $folder/prod
+    chmod -R 774 $folder
+    log_info "sudo permission needed to chown newly created '$folder'"
+    sudo chown -R :root $folder
```
J3y0, line 30: I think I am missing something but why do we need to `chown` by root ?

J3y0 (commented):
I am honnestly not qualified and have no idea of how to configure a `docker-compose.yml` file. I have no idea how it is constructed and as it has to be merged quickly, I have a plain confidence in what you have written ! Same thing for the `nginx`-related files. 

For the `Dockerfile`, this looks good to me ! Same for bash scripts, I just have one or two questions.

Finally, the README.md is very clear. I just add a suggestion because I was wondering myself how to check the status for a second and I would have love not to think about it and blindly copy the command right from the `README.md` !

Thanks a lot for the amazing PR. That will ease a lot the deployment !

scripts/setup-db-folder.sh
ctmbl: really good question!
> **Note**
> however we don't really `chown` it, we only change the group ownership tks to the `:`
> maybe I should have used `chgrp` instead of `chown`

This directory, `mongodb` and its subdirs `dev` and `prod` will be mounted in the containers so that the DB is persistent when the container is killed and restarted. You also need to know that when a file/dir is mounted in a container its UID/GID and permissions are kept.

Reading the bitnami image (the one we use) [doc](https://docs.bitnami.com/tutorials/bitnami-best-practices-hardening-containers/) (section `Use arbitrary UUIDs`) you can learn that they are **non-root images**
Meaning that the process inside is run as an unprivileged user *(the 1001 precisely, this obviously adds a good security layer and made me open #98)*, so he won't be able to read/write our mounted volume :confused: 
That's why it **belongs to the root group**, and why we need to change our mounted folder permissions: so that the mongdb process inside the container can write into this directory!

scripts/setup-db-folder.sh
ctmbl: > Are you sure you check the right path here ?

yes I am! Because in fact this test is also used in `gpg-share.sh` and `bump.sh` scripts and has been introduced by #29 so it has been tested :wink:
the reason why it seems not to work in your exmaple is because the real code is
```bash
root="$(dirname $(realpath $(dirname "$0")))"
```
when you used 
```
dirname $(realpath $(dirname .))
```
However in your case `. == /home/math/Documents/iscsc.fr` when `$0 == /home/math/Documents/iscsc.fr/scripts/setup-db-folder.sh`
That explains the differences experienced!

Just try:
```
cd /home/math/Documents/iscsc.fr/scripts
./setup-db-folder.sh
```
and you'll see the error message!

Finally you're right that's kind of weird but this is currently the way we're doing it :confused: 
maybe we should:
 1. factorize some `log*` and `check_*` functions used in all `*.sh` scripts
 2. find a better way to get the parent of the current directory (because that's what we're checking here!)

gbrivady (approved):
✔️

scripts/setup-db-folder.sh
J3y0: That's what I thought ! Thanks for the clarification. And for the other way, I was just wondering. This is details and we have to merge this, so don't take it into account for now

scripts/setup-db-folder.sh
J3y0: Thanks a lot for your response ! Makes sens to me now. I resolve this thread

J3y0 (approved):

PR #88, opened 2023-04-12
Refactor backend docker environment towards a python flask backend

ctmbl:
I simply refactored the `Dockerfile` and replace any use of `node-*` by `backend-*` or `flask-*`, for example for container name, env variable etc, I also added a `requirements.txt` for obvious reasons.

**To go further:**
I chose `flask` as future python backend framework out of [these propositions](https://www.monocubed.com/blog/top-python-frameworks/)

Two parts of a DigitalOcean `flask` tutorial, we could find useful:
 - setup a basic`flask` app [part 1/13](https://www.digitalocean.com/community/tutorials/how-to-create-your-first-web-application-using-flask-and-python-3#step-2-creating-a-simple-application) 
 - integrate MongoDB to `flask` [part 8/13](https://www.digitalocean.com/community/tutorials/how-to-use-mongodb-in-a-flask-application#prerequisites)

Currently the only thing that can be done is run the `docker-compose` command described in the `README.md`, **without certificates** (raw HTTP, after reverting #34 with `git revert 8087def` for example), and it builds and run without errors. Of course, none of the endpoints the frontend try to reach are currently implemented, it doesn't even use the right port, but this setup the environment to build upon!

We have 3 solutions described to configure the port: 
 - use of special `flask` env var [on StackOveflow](https://stackoverflow.com/a/55239694) or [on DigitalOcean's tuto](https://www.digitalocean.com/community/tutorials/how-to-create-your-first-web-application-using-flask-and-python-3#step-3-running-the-application) --> the one we should use according to me
 - run the app [from `app.py` in python](https://stackoverflow.com/a/55240457) (doesn't use `flask` CLI) --> reliable but seems like a hacky solution :shrug: 
 - [command line flags](https://stackoverflow.com/a/41940807) --> we shouldn't use that because it changes the docker image according to me

But all this will be work for other PRs to `backend/python`

:warning: IMPORTANT: i've read everywhere that the way I currently run the server (`flask run`) **is not** the proper way to do it in production, will have to investigate further! (for example [serve flask with Gunicorn and Nginx](https://www.digitalocean.com/community/tutorials/how-to-serve-flask-applications-with-gunicorn-and-nginx-on-ubuntu-20-04))


.env.example
```diff
@@ -3,7 +3,7 @@
 #####################
 
 REACT_PORT=3000  # local port for frontend
-NODE_PORT=3001  # local port for backend
+BACKEND_PORT=3001  # local port for backend
```
atxr, line 6: Not `FLASK_PORT`?
I think its better because the backend is a bit more than just the flask app, it's all the things which are running in the *back* (like flask and mongo)

atxr (commented):
Just a little nitpick :)

.env.example
ctmbl: You're right, but I'll change it to `FLASK_RUN_PORT` then because `flask` read it by default. It'll solve the PORT configuration choice I discussed in the Issue desc.

amtoine (approved):
lgtm :+1:

amtoine:
> **Note**
> once we land #91 and the `prettier` pipeline is fixed, you can
> ```bash
> git fetch iscsc
> git merge iscsc/main
> git push ctmbl
> ```
> to resolve the conflicts @ctmbl :+1:

ctmbl:
@amtoine ofc I'll do it, anyway the CI won't pass if I don't :+1:

PR #90, opened 2023-04-14
Fix CI issue: action fetches wrong repo

atxr:
This PR will solve the CI issue by adding the right repo to checkout.
Currently, when opening a PR from a fork, the prettier action fails to checkout the branch because it looks for it in iscsc/iscsc.fr instead of the forked repo.

I tested the fix with my fork.


amtoine (approved):
changes lgtm but i've no idea why this does not work in this repo :laughing: 

this is the very first and only time i've seen such a strange behaviour...
could there be a repo setting messing with us? :eyes:

PR #91, opened 2023-04-14
FIX: remove the `with` block under `actions/checkout@v3` in `prettier`

amtoine:
let's try to remove this `with` block under `actions/checkout@v3` in the `prettier` CI.


atxr (approved):
Perfect! Cleaner fix than mine :open_umbrella:

ctmbl (approved):

PR #92, opened 2023-04-14
Migrate database to localhost

atxr:
This PR removes the *hardcoded* values of the remote atlas database.
It introduces a new local database deployment in the docker-compose workflow.
The database is secured thanks to docker networking.

I tested the whole migration :heavy_check_mark: 

:warning: This PR doesn't migrate the current atlas db to the new local one. This action must be done "by hand" when the website will be bumped.

PR #93, opened 2023-04-17
Merge main changes to `backend/python`

ctmbl:
one PR in particular needs to be in `backend/python`: it's the recent #92


amtoine:
just to avoid a bad merge until we figure out how to do it :wink:

PR #95, opened 2023-04-20
Fix database container

### LINKED ISSUE #94: MongoDB container can't be reach and isn't persistent
**Problem:**
2 flaws were introduced by #92 (that should have been detected in review ooops):
 - the database isn't persistent: the wrong folder is mount
   we use a `bitnami/mongodb` image and according to the [documentation](https://hub.docker.com/r/bitnami/mongodb) the 
   volume to mount is `/bitnami/mongodb` not `/data/db`
 - the backend can't reach the DB **if** `DB_PORT != 27017` which kind of destroy the interest of having a config var...

**Solution:**
I found a solution to both of these problems in the documentation linked above (the right volume to mount and the `MONGODB_EXTRA_FLAGS` var to declare the right port) and I'll soon open a PR

*Additionally* the image 4.4 is quite old... we should agree on a newer version, I don't like using `latest` (but I'll use it if you have good arguments)
@atxr @amtoine

ctmbl:
close #94 

We should however take care of choosing an updated version of `bitnami/mongodb` image, 4.4 is pretty old... https://hub.docker.com/r/bitnami/mongodb

:warning: **IMPORTANT** from [bitnami documentation](https://hub.docker.com/r/bitnami/mongodb)
> NOTE: As this is a non-root container, the mounted files and directories must have the proper permissions for the UID 1001.

meaning `rw` I think (but maybe `x` too didn't test it)
This is part of this [debate](https://docs.bitnami.com/tutorials/work-with-non-root-containers/) and I think we should take greater care to use unprivileged users in our images... we should open an issue about that
At the moment it only means that when deploying we must ensure that this condition above is true, but a deployment script could take care of this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ certbot
 node_modules
 package-lock.json
 team.txt
+mongodb/**
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.8"
 
 services:
   mongodb:
-    image: docker.io/bitnami/mongodb:4.4
+    image: docker.io/bitnami/mongodb:latest
     networks:
       - database
     restart: always
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,8 @@ services:
       - database
     restart: always
     env_file: ./.env.production
+    environment:
+      - "MONGODB_EXTRA_FLAGS=--port=$DB_PORT"
     ports:
       - $DB_PORT:$DB_PORT
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,7 @@ services:
     ports:
       - $DB_PORT:$DB_PORT
     volumes:
-      - "./mongodb:/data/db"
+      - './mongodb/prod:/bitnami/mongodb'
 
   node-app:
     depends_on:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     ports:
       - $DB_PORT:$DB_PORT
     volumes:
-      - './mongodb/prod:/bitnami/mongodb'
+      - "./mongodb/prod:/bitnami/mongodb"
 
   node-app:
     depends_on:


atxr (commented):
Thanks for solving these issues I introduced earlier
I didn't approve yet because we need to tackle the bitnami/mongodb docker version before
Personally, I don't see any problem with `latest`, but if you want to choose a fixed version, let's update to 6.0 at least!

ctmbl:
sorry for the force push I add foolishly pushed a wrong commit (you know the revert of #34 to deploy locally with docker while #86 is not merge ahah)

gbrivady:
Works with the `npm` running method, commits and discussions/decisions you discutted seems sensible to me, so good to go

ctmbl:
@gbrivady yeah I knew that but I think it's time to change it so I promoted you to the `iscsc.fr-maintainers` team now your approval will allow this PR to be merged!

However, how did you start the DB if you've tried it with the npm command?

Also one important think you must test is the DB persistence, to properly test it you must also properly setup your `./mongodb` folder, don't know if you've already done it but here are the steps (implicitly explained in my first message):
```
mkdir -p mongodb/prod
mkdir mongodb/dev
chmod -R 664 mongodb
sudo chown -R :1001 mongodb
```
The goal of these lines is to make the mongodb folder (which will be mounted so shared by the container) owned and writable by the user 1001 in the container!

gbrivady:
> @gbrivady yeah I knew that but I think it's time to change it so I promoted you to the `iscsc.fr-maintainers` team now your approval will allow this PR to be merged!
> 
> However, how did you start the DB if you've tried it with the npm command?
> 
> Also one important think you must test is the DB persistence, to properly test it you must also properly setup your `./mongodb` folder, don't know if you've already done it but here are the steps (implicitly explained in my first message):
> ```
> mkdir -p mongodb/prod
> mkdir mongodb/dev
> chmod -R 664 mongodb
> sudo chown -R :1001 mongodb
> ```
> The goal of these lines is to make the mongodb folder (which will be mounted so shared by the container) owned and writable by the user 1001 in the container!
> 
> PS: if you'd like to test it properly with docker, without #86 being merged the best way is to :
> ```
> git revert 8087def
> ```
> unluckily you'll have a conflict, to solve it:
>  - in `docker-compose.yml` keep only the `network` section change, get rid of everything else, it's about SSL certifcates
>  - other conflicting files aren't code so add them as is we don't matter
> ```
> git revert --continue
> ```
> the you can start the docker
> ```
> docker-compose --env-file .env.production up --build
> ```
> you may have some env var wrongly setup but it's easily solvable

Yeah I might definitely have not checked it correctly, will try again tomorrow

gbrivady (approved):

ctmbl:
perfect! tks a lot for testing it!

PR #112, opened 2023-06-05
Add limit req to nginx conf to prevent DDOS

dolphinau:
In order to prevent DDOS attacks, I added in this PR a limit_req to the nginx production template.

### DIFF
--- a/nginx/nginx.conf.prod.template
+++ b/nginx/nginx.conf.prod.template
@@ -15,6 +15,9 @@ http {
     include mime.types;
     default_type application/octet-stream;
 
+    # Avoid DDOS
+    limit_req_zone ${DOLLAR}binary_remote_addr zone=one:10m rate=5r/s;
+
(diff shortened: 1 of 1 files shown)


ZynoXelek (approved):

nginx/nginx.conf.prod.template
ctmbl, line 1: Changes look good but couldn't we define `limit_req` at a more global level to avoid this config duplication? This seems annoying to maintain and very error prone :confused:

ctmbl (commented):
Looks good but we need to really test it before approving!

nginx/nginx.conf.prod.template
dolphinau: Actually I don't know, I didn't manage to find any documentation where the limit req was "global"

ctmbl: Ok never mind then it's no big deal!

